PR #8, opened 2018-06-26
build: Insert copyright in cmake files

mikew-lunarg:


karl-lunarg:
For others who are curious, the three-tilde fences are to keep the CMake source code formatter out of the comment text.

PR #17, opened 2018-08-06
Add some missing "optional" markers

Ralith:
The comments on these members suggest that it is sometimes legal for them to contain `VK_NULL_HANDLE`, or in some cases an undefined value. Unless I've misunderstood the schema, it seems like they should therefore be marked as optional.


karl-lunarg:
@Ralith  You'll probably want to move this PR to KhronosGroup/Vulkan-Docs.  See the README in this repo.

PR #24, opened 2018-10-10
headers: Update to Vulkan header version 1.1.87

shannon-lunarg:
Updated:
- include/vulkan/vulkan.h
- include/vulkan/vulkan.hpp
- include/vulkan/vulkan_core.h
- registry/genvk.py
- registry/reg.py
- registry/validusage.json
- registry/vk.xml

Added:
- include/vulkan/vulkan_fuchsia.h

Note: A local, modified version of Vulkan-Docs/xml/reg.py was used to
generate these files. This modification filters out disabled extensions
when populating the structextends list for a given base struct.

### DIFF
--- a/include/vulkan/vulkan.h
+++ b/include/vulkan/vulkan.h
@@ -24,6 +24,10 @@
 #include "vulkan_android.h"
 #endif
 
+#ifdef VK_USE_PLATFORM_FUCHSIA
+#include <zircon/types.h>
+#include "vulkan_fuchsia.h"
+#endif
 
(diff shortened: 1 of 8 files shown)


oddhack:
Please propagate the script fix back to the gitlab vulkan repo. I expect we'll run into this again at some point (and it is the right thing to do for other reasons).

mikes-lunarg:
> Please propagate the script fix back to the gitlab vulkan repo. I expect we'll run into this again at some point (and it is the right thing to do for other reasons).

https://gitlab.khronos.org/vulkan/vulkan/merge_requests/2864

PR #27, opened 2018-10-30
Compare to `None` using identity `is` operator

janisozaur:
This is a trivial change that replaces `==` operator with `is` operator, following PEP 8 guideline:

> Comparisons to singletons like None should always be done with is or is not, never the equality operators.

https://legacy.python.org/dev/peps/pep-0008/#programming-recommendations

### DIFF
--- a/registry/cgenerator.py
+++ b/registry/cgenerator.py
@@ -189,7 +189,7 @@ class COutputGenerator(OutputGenerator):
             # If type declarations are needed by other features based on
             # this one, it may be necessary to suppress the ExtraProtect,
             # or move it below the 'for section...' loop.
-            if (self.featureExtraProtect != None):
+            if (self.featureExtraProtect is not None):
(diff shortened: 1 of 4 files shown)


shannon-lunarg:
The Vulkan-Headers repo is different from other related repos, e.g., Vulkan-ValidationLayers or LunarG-VulkanSamples, in that the scripts found in the `scripts` directory in this repo are directly copied from the `xml` directory of the [Vulkan-Docs](https://github.com/KhronosGroup/Vulkan-Docs) repo. My suggestion would be to open a PR with these modifications in the Vulkan-Docs repo so that Vulkan-Headers will eventually pick up these changes (assuming the PR is accepted).

Thanks for your contributions!

oddhack:
Closing this. Pending updates from the 1.1.105 Vulkan-Docs spec update include this and many other PEP8 changes to the scripts.

PR #28, opened 2018-11-06
headers: Update to Vulkan header version 1.1.91

shannon-lunarg:
Updated:
- `include/vulkan/vulkan.h`
- `include/vulkan/vulkan.hpp`
- `include/vulkan/vulkan_core.h`
- `registry/genvk.py`
- `registry/validusage.json`
- `registry/vk.xml`

### DIFF
--- a/include/vulkan/vulkan.h
+++ b/include/vulkan/vulkan.h
@@ -39,12 +39,6 @@
 #endif
 
 
-#ifdef VK_USE_PLATFORM_MIR_KHR
-#include <mir_toolkit/client_types.h>
-#include "vulkan_mir.h"
-#endif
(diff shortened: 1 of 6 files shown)


mark-lunarg:
@shannon-lunarg, if you've not already taken note, the genvk.py changes should get folded into lvl_genvk.py when the KG is updated, if not before.

ewerness-nv:
I have a change ready to make validation build with this header change, but was holding it until the header change was ready: https://github.com/ewerness-nv/Vulkan-ValidationLayers/commit/e650ad700cd5ad04e675b1eb0fc5b5f95069bdd1

PR #30, opened 2018-11-21
build: Fix install target path

nsubtil:
Use CMAKE_CURRENT_SOURCE_DIR instead of CMAKE_SOURCE_DIR when defining
the path for the source files to be installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,8 +36,8 @@ endif()
 
 # --------------------------------------------------------------------------------------------------------------------------------
 
-install(DIRECTORY "${CMAKE_SOURCE_DIR}/include/vulkan" DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
-install(DIRECTORY "${CMAKE_SOURCE_DIR}/registry" DESTINATION ${CMAKE_INSTALL_DATADIR}/vulkan)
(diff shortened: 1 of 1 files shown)


mikew-lunarg:
@nsubtil In the PR description, could you elaborate on the motivation for this change?

karl-lunarg:
This is needed when another CMake project includes this project via CMake's `add_subdirectory()`.  Since there is no "target" associated with these files, they have to be explicitly provided to the `install()` command.  Installed files coming from actual CMake targets usually don't have to be called out this way.

And since the files need to be explicitly provided, that is done via a reference to the source code location.  `CMAKE_SOURCE_DIR` is where CMake was invoked and that would be up in the project that contains this one.  The header files are not there.  `CMAKE_CURRENT_SOURCE_DIR` correctly locks into the source tree in the Vulkan-Headers project and finds the files relative to the Vulkan-Headers tree.

Note that this won't have any effect when building Vulkan-Headers standalone since `CMAKE_SOURCE_DIR` == `CMAKE_CURRENT_SOURCE_DIR` in that case.

mikew-lunarg:
@karl-lunarg Ok
@nsubtil Something like that explanation should make its way into the commit message.

nsubtil:
Sorry for the delay. @karl-lunarg's comment is spot on.

I added a brief explanation of why this is needed to the commit message.

mikew-lunarg:
@nsubtil Thanks, pushed to master

PR #31, opened 2018-11-21
headers: Update to Vulkan header version 1.1.93

Tobski:
Needed for validation work for EXT_scalar_block_layout - hopefully I've done this right...?


Tobski:
@shannon-lunarg I think I actually missed the valid usage json file now I look over the wiki again - figuring out how to build vulkan-hpp correctly took too many cycles I guess! Anyway, if you're updating now, I'll close this :)

PR #34, opened 2018-12-11
headers: device creation callbacks for layers

felixdoerre:
Header changes for https://github.com/KhronosGroup/Vulkan-Loader/issues/74


lenny-lunarg:
I posted a review for the loader changes in the related [loader PR](https://github.com/KhronosGroup/Vulkan-Loader/pull/117). To see my feedback, look there. I would view these two PRs as a package, because there's no value to having this without the loader changes, So to see my feedback, look on the loader issue.

PR #41, opened 2019-02-08
build: Define library targets for nested builds

nsubtil:
Adds alias interface libraries Vulkan::Headers and Vulkan::Registry that
export include paths. These are meant to be used by dependent projects
in a nested build configuration to obtain the path to the Vulkan headers
and registry directories.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,15 @@ endif()
 
 # --------------------------------------------------------------------------------------------------------------------------------
 
+# define exported targets for nested project builds to consume
+add_library(Vulkan-Headers INTERFACE)
+target_include_directories(Vulkan-Headers INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/include")
+add_library(Vulkan::Headers ALIAS Vulkan-Headers)
+
+add_library(Vulkan-Registry INTERFACE)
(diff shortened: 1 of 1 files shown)


nsubtil:
@Tony-LunarG @mikew-lunarg @karl-lunarg this is a prototype of the changes I mentioned in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/639.

PRs https://github.com/KhronosGroup/Vulkan-Loader/pull/142 and https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/682 are the corresponding Loader and VL changes.

These patches may still need a few changes, but I'd like to get an initial reaction to see if this looks usable.

karl-lunarg:
FWIW, I think that this is a good direction.  This helps things like the validation layers that need the headers, but not the library.

mikew-lunarg:
LGTM  @lenny-lunarg may I ask you to review and merge?  
I'm OoO next week, and it's bad practice to merge then leave for an extended period.

mikew-lunarg:
Need to run this on internal CI farm.

mikew-lunarg:
Pushed to master at commit 2ed3359a526b808cc6f06115e2e55bb03699d072

PR #44, opened 2019-03-04
Update for Vulkan-Docs 1.1.102

oddhack:
@mark-lunarg @shannon-lunarg @nsubtil @mtavenrath this is my first attempt to move the Vulkan-Headers repo update into my weekly Vulkan-Docs spec update process, per internal issue 745. Please review and, preferably, try to reproduce against the Vulkan-Docs v1.1.102 tag, then accept the PR if it looks OK. I believe I have hit all the required points here

If this looks OK then I'll proceed to the next stage, removing the generated headers from the Vulkan-Docs tree, in the next spec update cycle. After that hopefully we'll have gained enough confidence that I can just directly merge into master once


mark-lunarg:
This is great! Looks fine to me, but @shannon-lunarg should have our final approval.

nsubtil:
This does look like a good step to me as well. Not qualified to comment on the PR itself, but happy with the general direction.

FWIW, I have been planning to add a CMake target to the Vulkan-Docs repo to generate the headers, for better integration with the nested build.

oddhack:
@shannon-lunarg OK, I'll go ahead and merge, and be responsible for this step going forward. When we run into build problems at some point in the future, I'll just leave the branch in the tree and raise an issue here or with Vulkan-Hpp to resolve it; otherwise will just merge the updates myself, if that's OK.

mark-lunarg:
Thanks @oddhack -- perfectly fine for you to merge if there are no apparent issues. This is a nice process improvement!

PR #47, opened 2019-03-17
Update for Vulkan-Docs 1.1.104

oddhack:
@shannon-lunarg @mtavenrath as previously noted, there's a problem generating Vulkan-Hpp for this week. I filed https://github.com/KhronosGroup/Vulkan-Hpp/issues/303 and suggest this PR be held until that's fixed and this branch is updated - there are a bunch of new extensions in the 1.1.104 update.

There are many benign whitespace changes in the generated C headers, due to scripting changes.

Also, there were a lot of script changes in v1.1.104, which may affect downstream consumers like the LVL repos. Here I just needed to add the two new *conventions.py scripts.

PR #48, opened 2019-03-20
Update for Vulkan-Docs 1.1.105

oddhack:
This supplants #47. But has the same general issue - there's another Vulkan-Hpp build problem related to new extensions this week, see https://github.com/KhronosGroup/Vulkan-Hpp/issues/309 .

I don't know the full set of intended use cases for this repo, but there have been 2 spec updates containing 7 new extensions published since the last Vulkan-Headers update, some of them probably of significant interest due to the GDC announcement. So ISTM we ought to update the C headers and XML/scripts via this PR, even though we can't update Vulkan-Hpp yet. @shannon-lunarg let me know what you think.


pdaniell-nv:
This header update won't compile on Windows because the definition for VkPhysicalDeviceSurfaceInfo2KHR appears in both vulkan_core.h and vulkan_win32.h. Perhaps there is a coverage gap in CI.

oddhack:
I pushed a commit to Vulkan-Headers which fixes the double definition problem. I will setup a compilation test with fake platform headers so CI can compile all platforms simultaneously, and catch this sort of thing in the future (until someone adds a new platform and forgets to add it to the CI test).

I'm still waiting for a Vulkan-Hpp fix before progressing this, but this branch will pick up the C header update when I do.

oddhack:
@pdaniell-nv @shannon-lunarg I believe that both C and C++ headers are working properly now, thanks to help from the Vulkan-Hpp folks. Since there have been multiple issues with this, I'd appreciate some oversight / signoff from you folks before merging this PR, however.

oddhack:
OK. I've addressed @pdaniell-nv's issue so I won't wait on his signoff.

PR #50, opened 2019-04-05
[WIP] registry: Disable `VK_AMD_display_native_hdr` ext

shannon-lunarg:
The `VK_AMD_display_native_hdr` extension introduced the
`SetLocalDimmingAMD` API command which takes a nondispatchable type as
its first parameter, resulting in failures for 32-bit Windows builds.


oddhack:
This will be fixed in the next Vulkan-Docs update, planned for Monday. AMD is adding a VkDevice parameter to the command. So if you don't require a fix before then, I recommend closing this and waiting for me to do the public update.

oddhack:
Closing - about to accept 1.1.106 update, which adds a dispatchable parameter in vk.xml.

PR #56, opened 2019-05-08
build: Update cmake_minimum_required(VERSION 3.12.4)

mikew-lunarg:
Update both CMakeLists.txt files and documents to the new required CMake version.
This PR will not be merged to master until the CI systems are updated.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -78,7 +78,7 @@ directories and place them in any location.
     - [2017](https://www.visualstudio.com/vs/downloads/)
   - The Community Edition of each of the above versions is sufficient, as
     well as any more capable edition.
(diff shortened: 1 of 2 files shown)


oddhack:
I'm OK with this - not too familiar with cmake, but the required version is lower than that shipped with Debian Jessie, so it shouldn't cause me any issues.

mikew-lunarg:
Thanks.  What is your CMake version on Debian?

oddhack:
Actually I'm incorrect - I'm running off stretch-backports, which has 3.13.2, but stretch itself just comes with 3.7.2. From my POV this is OK, but I take Jesse's point in email about organizations with older toolchains.

PR #68, opened 2019-07-24
Fix typo.

orbea:
I think this is supposed to be `__clang__` and not `__clang___` as seen in elsewhere in the repo and in online documentation.

### DIFF
--- a/include/vulkan/vulkan.hpp
+++ b/include/vulkan/vulkan.hpp
@@ -100,7 +100,7 @@ static_assert( VK_HEADER_VERSION ==  116 , "Wrong VK_HEADER_VERSION!" );
 #endif
 
 #if !defined(VULKAN_HPP_INLINE)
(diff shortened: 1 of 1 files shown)


oddhack:
@orbea vulkan.hpp is a completely autogenerated file, so to make changes in it you'll need to propose a change to the generator code, which is in the https://github.com/KhronosGroup/Vulkan-Hpp/ repository. I suggest closing this PR and filing an issue in Vulkan-Hpp; once the maintainers deal with it, it will take effect here the next time we push a spec and header update.

orbea:
Thanks for the explanation, I will bring this up in the https://github.com/KhronosGroup/Vulkan-Hpp/ repository.

orbea:
New PR made here. https://github.com/KhronosGroup/Vulkan-Hpp/pull/365

PR #70, opened 2019-07-29
Compatible pathlib for py2 and py3 in registry/generator.py

haoxli:


oddhack:
All content in this repository is generated based on the Vulkan-Docs and Vulkan-Hpp repositories. In this case you should retarget the PR to the scripts in Vulkan-Docs.

PR #73, opened 2019-08-12
Update for Vulkan-Docs 1.1.119

oddhack:
Vulkan-Hpp does not build, yet.


oddhack:
vulkan.hpp builds and compiles now.

PR #84, opened 2019-10-25
build: BUILD.gn missing defines for MacOS

jonahryandavis:
BUILD.gn does not define VK_USE_PLATFORM_MACOS_MVK for MacOS

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -28,6 +28,9 @@ config("vulkan_headers_config") {
   if (is_fuchsia) {
     defines = [ "VK_USE_PLATFORM_FUCHSIA" ]
   }
+  if (is_mac) {
+    defines = [ "VK_USE_PLATFORM_METAL_EXT" ]
+  }
 }
 
 # Vulkan headers only, no compiled sources.


jonahryandavis:
@mikes-lunarg Can I ask you to take a look at this additional patch related to ANGLE+SwiftShader for Mac?

BUILD.gn
```diff
@@ -28,6 +28,9 @@ config("vulkan_headers_config") {
   if (is_fuchsia) {
     defines = [ "VK_USE_PLATFORM_FUCHSIA" ]
   }
+  if (is_mac) {
+    defines = [ "VK_USE_PLATFORM_MACOS_MVK" ]
```
mikes-lunarg: Does this have to be the MVK define? I thought Vulkan implementations on macOS were headed toward VK_EXT_metal_surface/VK_USE_PLATFORM_METAL_EXT

jonahryandavis: Swiftshader implements the MVK one, but I just tested a patch to implement the EXT one and it works. I'll use the EXT one in this patch

PR #90, opened 2019-11-16
GN: Make PLATFORM_XCB optional, based on a user-defined variable.

flibitijibibo:
Not all Linux platforms will have X11 available (Wayland, GGP), so we have to declare a variable that checks use_x11 separately.

This depends on an ANGLE bug: https://bugs.chromium.org/p/angleproject/issues/detail?id=4116

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -13,13 +13,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import("//build_overrides/vulkan_headers.gni")
+
 config("vulkan_headers_config") {
   include_dirs = [ "include" ]
 
   if (is_win) {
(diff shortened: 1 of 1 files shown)


flibitijibibo:
The relevant patch for ANGLE has been merged, so this is now ready to review:

https://chromium.googlesource.com/angle/angle/+/d22b8be4b457951ac255991d300845199753e063

flibitijibibo:
Checking in on this - are there any possible blockers for this I can look at?

KarenGhavam-lunarG:
I believe folks were waiting for @oddhack to approve and merge.

oddhack:
I don't actually even know what this file does, but since you folks have signed off on it that's fine with me - my only part in this repo is populating it with headers generated from Vulkan-Docs.

PR #99, opened 2020-01-17
Fix incorrect alias name vkCmdDrawIndexedIndirectCountKHR

guchong1988:


oddhack:
Wrong repository - Vulkan-Headers just serves to assemble generated artifacts & supporting material from other repos. Should be refiled against Vulkan-Docs which is where the canonical vk.xml lives. Also, please leave the currently-unreachable FlagBits type, it's there for future use. Thanks.

PR #104, opened 2020-03-13
headers: Simplify Non-Dispatchable Handles

awr1:
Making the ND handles a struct containing a 64-bit value should make the preprocessor branch/pointer defintion unnecessary and also allow the strong typedef to exist on 32-bit platforms.

Related: https://stackoverflow.com/questions/51743500/why-do-non-dispatchable-handles-use-a-ptr-on-64bit-platofrms

### DIFF
--- a/include/vulkan/vulkan_core.h
+++ b/include/vulkan/vulkan_core.h
@@ -54,15 +54,7 @@ extern "C" {
 
 
 #define VK_DEFINE_HANDLE(object) typedef struct object##_T* object;
-
-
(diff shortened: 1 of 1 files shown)


mark-lunarg:
@awr1, this repository is downstream of the main Vulkan repository which defines header content, and so this issue should instead be raised there, in [Vulkan-Docs](https://github.com/KhronosGroup/Vulkan-Docs).

awr1:
Thanks @mark-lunarg, I will divert this over to there.

PR #110, opened 2020-04-12
Add CMake install target

NeroBurner:
continuation and rebase of #89 from @rbsheth

I also added that the VERSION_STRING is read from `vulkan_core.h`

If wanted I can squash the commits


mikew-lunarg:
Hey @NeroBurner 
We're thinking along similar lines...

In the background I've been working on a script to be a centralized extractor of version information from the Vulkan headers; see
https://github.com/mikew-lunarg/vulkan_utils/blob/master/VulkanVersionUtils.py

The idea is to factor out all the different implementations of version extraction (from code/json generators and CMake files) and have one tool everyone can call that handles all the regex and output formatting.

It's not quite ready yet; hope to have it in the Vulkan repos soon.

May I ask you to re-imagine this PR in terms of that script?

Thanks
-- mew

NeroBurner:
My target is to have a versioned config.cmake installed with the headers. The current PR needs nothing else, but cmake to do this. With the python script one would need python as a build dependency. Python is harder to manage system wide in windows (a horror in my oppinion). For CI builds python would be another dependency.
I think this is a hefty price to pay, if the cnake-only solution is already ready

The cmake solution would also make the custom FindVulkanHeaders.cmake script in the VulkanLoader package obsolete, and it could be removed

mikew-lunarg:
@oddhack a thought...

Perhaps an easily grep-able string representation of the dotted-triple version could be added as a comment to the header as part of the generation from XML?  That would not expand the C API.

That would avoid some of the regex gymnastics to extract the version from the header for non-C tools.

NeroBurner:
An easier grep-able version comment is not in conflict with installing the cmake config file. I'd like to propose to merge this PR and if need be I'd update the cmake portion after the generator change

mark-lunarg:
Anybody tried this with the downstream repos?

NeroBurner:
Just tried with Vulkan-Loader, which has a `FindVulkanHeaders.cmake` file. After deleting it Vulkan::Headers was not found anymore, because this PR creates the config `${PROJECT_NAME}Config.cmake`, which translates to `Vulkan-HeadersConfig.cmake`. This would require a call to `find_package(Vulkan-Headers)`. Changed that with the last commit

Vulkan::Loader changed the subdirectory-adding with the commit https://github.com/KhronosGroup/Vulkan-Loader/commit/ba1f7760c20e2925d71b4a5783fd4c34c7c2e980

Now the target `Vulkan::Headers` is expected before it even has a chance to be added. If  we add the line `find_package(VulkanHeaders CONFIG)` just before the `if (TARGET Vulkan::Headers)` we could use the newly generated config file from this PR, and fall back to the `FindVulkanHeaders.cmake` otherwise

mark-lunarg:
@NeroBurner, thanks for checking Vulkan-Loader.  There are almost half-a-dozen others we'll need to check (I'll sign up for validation layers) before this can land, and I think it'd be a good idea to have @karl-lunarg weigh in, if he can spare the time.

NeroBurner:
@mark-lunarg if possible, could you post a list of projects we need to check? Then we could check them off one by one

CMakeLists.txt
```diff
@@ -22,6 +22,16 @@ cmake_minimum_required(VERSION 3.10.2)
 
 # NONE = this project has no language toolchain requirement.
 project(Vulkan-Headers NONE)
+file(READ "include/vulkan/vulkan_core.h" ver)
+string(REGEX MATCH "#define VK_HEADER_VERSION_COMPLETE VK_MAKE_VERSION\\(([0-9]+), ([0-9]+), VK_HEADER_VERSION\\)" _ ${ver})
```
lunarpapillo, line 26: I'm a little worried that these regexes are whitespace-sensitive.  I believe it's legal to have whitespace between `#` and `define`, and an error as minor as an extra space after `VK_HEADER_VERSION_COMPLETE` would not affect compilation but would break this process.

CMakeLists.txt
```diff
@@ -22,6 +22,16 @@ cmake_minimum_required(VERSION 3.10.2)
...
+set(VERSION_MAJOR "${CMAKE_MATCH_1}")
+set(VERSION_MINOR "${CMAKE_MATCH_2}")
+string(REGEX MATCH "#define VK_HEADER_VERSION ([0-9]+)" _ ${ver})
+set(VK_HEADER_VERSION "${CMAKE_MATCH_1}")
+set(VERSION_STRING "${VERSION_MAJOR}.${VERSION_MINOR}.${VK_HEADER_VERSION}")
```
lunarpapillo, line 31: Some of the values we're setting are protected with an initial `VK_` (`VK_HEADER_VERSION`) while others are not (`VERSION_MAJOR`, `VERSION_MINOR`, `VERSION_STRING`).  Is there a reason they're not all protected with `VK_`?

mikew-lunarg:
@NeroBurner LunarG is testing this PR with all our downstream repos.  Assuming all goes well, I'll merge.

mark-lunarg:
VVL is OK.

jeremyk-lunarg:
V-T built and ran with both CMake 3.10.2 and the newest CMake.

lenny-lunarg:
Loader is good.

Tony-LunarG:
Samples is good

mikew-lunarg:
@NeroBurner while we're testing, would you squash the commits and enhance the description of what's contained.

davidlunarg:
VulkanTools builds

CMakeLists.txt
NeroBurner: improved the regex to be more robust. I left out the check for space between `#` and `define`, and I hope that no-one will use this capability.

I also added checks to see if one of the regex did not match and write a `FATAL_ERROR`

CMakeLists.txt
NeroBurner: no reason at all, added the `VK_` guards

NeroBurner:
@mikes-lunarg squashed, cleanup indentations, fixed that only one include path is in the interface include property instead of two times the same path, checks for version-parsing

Please tell me if something is unclear and needs an explaining comment

lunarpapillo:
I double-checked all five test runs, and examined one Linux and one Windows run from each in more detail.  All passed.  In addition, I verified for the **Vulkan-ValidationLayers** runs that the new CMake message `Found Vulkan-Headers VERSION_STRING` did *not* appear in the `cmake` output (as expected, because we're locked at CMake 3.10.2).

mikew-lunarg:
We have two mikes, I'm @mikew-lunarg  :-)

mikew-lunarg:
@NeroBurner  in the description, if you would add a list of all three new files that get added to $INSTALL_PREFIX, I think we're set.

NeroBurner:
added to commit description and in the CMakeLists.txt file as well

NeroBurner:
You're very welcome! :tada:

rbsheth:
Thanks all, and @NeroBurner for helping #89 get in!

mark-lunarg:
@NeroBurner, can you please take a look at [Vulkan-ValidationLayers issue #1752](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/1752)?

mikew-lunarg:
@NeroBurner Due to this PR #110 breaking downstream repos, and proposed fixes to this PR from #112, we're going to revert this PR to undo the breakage.

Further discussion and development should occur on #112

cc @dj2

PR #112, opened 2020-04-23
[WIP] Cleanup CMake install target

NeroBurner:
second try of https://github.com/KhronosGroup/Vulkan-Headers/pull/110

Also removed the extra complication of creating alias over alias for the install targets. Now cmake itself can generate Vulkan::Headers at install. Less complicated, less chance to break

Changes on downstream-repos of Vulkan-Headers are probable, mostly because those downstram projects use custom FindVulkanHeaders and then create a Vulkan::Headers target without checking if the target already exists

need to check downstream projects:

- https://github.com/KhronosGroup/Vulkan-Loader/pull/380
- https://github.com/KhronosGroup/Vulkan-Tools/pull/382
- https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/1755
- https://github.com/LunarG/VulkanSamples/pull/290
- https://github.com/LunarG/VulkanTools/pull/994
- [google/amber](http://github.com/google/amber)
- [google/vkb](http://github.com/google/vkb)

edit: add amber and vkb downstream repos to check, recommended by KarenGhavam-lunarG
edit2: add related PRs and Vulkan-ValidationLayers project


mikew-lunarg:
@NeroBurner PR #110 has been reverted, please rebase this PR.

NeroBurner:
The alias for add_subdirectory is still there. 

I just renamed the original Target from VulkanHeaders to Headers. When installing cmake then creates the Imported and namespaced Vulkan::Headers target without having to define an Alias in VulkanHeadersConfig.cmake

NeroBurner:
rebased

KarenGhavam-lunarG:
Your list of known downstream repos that might be impacted:

> KhronosGroup/Vulkan-Loader
> KhronosGroup/Vulkan-Tools
> LunarG/VulkanSamples
> LunarG/VulkanTools

I believe these repositories also might be impacted and need to be evaluated:
Both http://github.com/google/amber and http://github.com/google/vkb pull in and build the loader and layers.

NeroBurner:
@KarenGhavam-lunarG thank you, updated the original list

rbsheth:
@NeroBurner Is this still WIP? I'm looking into working on this again now.

NeroBurner:
@rbsheth unfortunately yes, the PR itself should be in a good state, but the downstream projects would need to be updated in the same swoop. But due to time constraints I did not finish those, so this PR is not finished

- https://github.com/KhronosGroup/Vulkan-Tools/pull/382
- https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/1755
- https://github.com/LunarG/VulkanTools/pull/994
- https://github.com/LunarG/VulkanSamples/pull/290

You're very welcome to work on this PR or the downstream PRs

dj2:
So, this will break downstream repos? Is it possible to set it up so both the old and new way works? Trying to identify all downstream users is going to miss and break some folks I'm guessing.

CMakeLists.txt
```diff
@@ -37,22 +57,82 @@ endif()
...
-install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/registry" DESTINATION ${CMAKE_INSTALL_DATADIR}/vulkan)
-
-# uninstall target
-if(NOT TARGET uninstall)
-    configure_file("${CMAKE_CURRENT_SOURCE_DIR}/cmake/cmake_uninstall.cmake.in"
-                   "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake"
-                   IMMEDIATE
-                   @ONLY)
-    add_custom_target(uninstall COMMAND ${CMAKE_COMMAND} -P ${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake)
+add_library(Headers INTERFACE)
```
SpaceIm, line 60: No need to rename the target (and breaks downstream projects). You can rely on https://cmake.org/cmake/help/git-stage/prop_tgt/EXPORT_NAME.html to export the target with another name.

KarenGhavam-lunarG:
This PR hasn't progressed in 18 months. I think it is appropriate to close this.

PR #135, opened 2020-07-22
Fix enum order for backwards compatibility

haasn:
The new enum member introduced in 2a164e968 broke ABI with previous
versions of vk_icd.h, resulting in a segfault on latest mesa (which does
not yet have this commit in their vendored copy of vk_icd.h, thus
breaking the order of enum members).


kodabb:
> @lenny-lunarg This makes me think it's likely a good idea to just specify the numeric values of those enum values to ensure that we don't break things by changing the order in the future. In any case, we should take this first.

I'd recommend #define statements instead of enums if you go that route, alternatively only adding new enums at the end of the list is fine ^^

PR #140, opened 2020-08-12
Add VI WSI platform support to vk_icd.h

mchock-nv:
Add VI enum value to VkIcdWsiPlatform and define type VkIcdSurfaceVi.

### DIFF
--- a/include/vulkan/vk_icd.h
+++ b/include/vulkan/vk_icd.h
@@ -120,6 +120,7 @@ typedef enum {
     VK_ICD_WSI_PLATFORM_HEADLESS,
     VK_ICD_WSI_PLATFORM_METAL,
     VK_ICD_WSI_PLATFORM_DIRECTFB,
+    VK_ICD_WSI_PLATFORM_VI,
 } VkIcdWsiPlatform;
 
 typedef struct {
@@ -216,5 +217,12 @@ typedef struct {
     const CAMetalLayer *pLayer;
 } VkIcdSurfaceMetal;
 #endif // VK_USE_PLATFORM_METAL_EXT
+
+#ifdef VK_USE_PLATFORM_VI_NN
+typedef struct {
+    VkIcdSurfaceBase base;
+    void *window;
+} VkIcdSurfaceVi;
+#endif // VK_USE_PLATFORM_VI_NN
 
 #endif  // VKICD_H


mchock-nv:
Tagging @lenny-lunarg

lenny-lunarg:
This change looks exactly as it should for this platform. My only question is if this means you'll also be looking for loader support. The [khronos loader](https://github.com/KhronosGroup/Vulkan-Loader) doesn't currently support this platform and we haven't had any plans to add that support. Android uses its own loader, so I've always just viewed VI as a similar case.

mchock-nv:
As I understand it, there are no current plans to add support to the Khronos loader.

Incidentally, there does seem to be something going on with the CLA checker; I signed the CLA hours ago, and then again a few minutes ago. I don't know whether my say-so here is an adequate substitute.

oddhack:
> Incidentally, there does seem to be something going on with the CLA checker; I signed the CLA hours ago, and then again a few minutes ago. I don't know whether my say-so here is an adequate substitute.

@outofcontrol could you check on the CLAbot - github integration? Seems like something has stalled.

mchock-nv:
Still stuck. I can get to a page that claims I've signed the CLA, but no change here.

Khronoswebmaster:
The CLA sporadically has issues which directly affects peoples ability to sign the CLA. There is at least one issue opened on the CLA website claiming an SSL issue is blocking the signing. If you are comfortable with the browser console, it would be useful to know if you are seeing errors which we can pass along the CLA Assistant developers.

mchock-nv:
Reloading the page or clicking the "recheck" link yields the warning

> Cookie “_ga” will be soon rejected because it has the “sameSite” attribute set to “none” or an invalid value, without the “secure” attribute. To know more about the “sameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite analytics.min.js:4:1234

Clicking the CLA link yields a warning about using the "zoom" attribute and some debug messages, but nothing that looks wrong to my non-web-engineer eyes.

oddhack:
I signed the CLA a couple of times yesterday in other repos so it's not a general infrastructure problem. @mchock-nv while it's a horrid "workaround", you could try another browser or a different browser profile (my main Chromium profile has more than a dozen active extensions and not infrequently has weird issues with some sites, I also keep Firefox handy for extra-challenging cases).

mchock-nv:
I had been using Firefox. I also just tried Incognito-mode Chrome, and iPad Safari off the company VPN. All three take me to a page that congratulates me on signing the CLA, but I see no change here. Given that I get to the "You have signed the CLA for KhronosGroup/Vulkan-Headers" page with the correct username visible on the same page, it seems like it has to be a server-side issue, though possibly specific to my account somehow.

lenny-lunarg:
@oddhack, if we can't get the CLA working are we going to be able to just merge this? It's good to stick to Khronos protocols and everything but it's been 5 days at this point and it still doesn't seem to be working.

Khronoswebmaster:
Looks like @kevin-mccullough needs to sign-off still, according to this commit: https://github.com/KhronosGroup/Vulkan-Headers/pull/140/commits/8ba5f6ecb8283d4e13e6f54c73ac9f8ecf0d1590

Khronoswebmaster:
@lenny-lunarg Unfortunately it is not a protocol issue, it is a legal issue. Once Kevin signs the CLA,  you should be able to commit.

lenny-lunarg:
@Khronoswebmaster, good point about the commit author.

@mchock-nv, it looks like this commit was authored by @kevin-mccullough. As a result, I think the CLA is working correctly because it's the actual author (not committer) that needs to sign it. If we can get the CLA signed by him then I think we should be able to merge.

mchock-nv:
Well, maybe that's the problem. I'm not @kevin-mccullough, but I am the committer for the change.

mchock-nv:
At Kevin's request I changed the author to myself, which clears the CLA issue.

oddhack:
@Khronoswebmaster would it be possible to associate github & Khronos internal IDs so that the CLAbot knows whenever the signer is someone who already falls under the Khronos member agreement?

outofcontrol:
@oddhack Is there a preferred action to take if we can make this association?

oddhack:
Check with Neil, but AIUI everyone who's subject to the Khronos member agreement has already assigned licensing rights to Khronos for their PRs. So that could be skipped, potentially, for github PRs coming from members. Definitely confirm with Neil, though.

PR #144, opened 2020-08-31
Introduce union for acceleration structure host handle and device addr

MarijnS95:
According to the [VkAccelerationStructureInstanceKHR docs] `accelerationStructureReference` takes an AS device address or VkAcelerationStructureKHR on the host. Wrap these in a union to clarify intent and aid in proper type usage in generated bindings.

Since a device address is specifically concerned in a device context, update `uint64_t` to `VkDeviceAddress` (which is a `typedef` to `uint64_t`).

[VkAccelerationStructureInstanceKHR docs]: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkAccelerationStructureInstanceKHR.html


dgkoch:
per https://gitlab.khronos.org/vulkan/vulkan/-/issues/2338 not doing this (we tried this before and it didn't work out)

PR #150, opened 2020-10-12
Update for Vulkan-Docs 1.2.157

oddhack:


Friz64:
Seems like this is missing the updates to vk.xml?

oddhack:
> Seems like this is missing the updates to vk.xml?

Thanks for catching that! Should be fixed now and the tag changed to point at the version with new vk.xml.

PR #155, opened 2020-10-20
Fix T & ArrayWrapper1D::operator[](int index)

xantares:
With MinGW I get the following error, because constexpr clashes with const:
vulkan.hpp:544:30: error: ‘constexpr T& vk::ArrayWrapper1D<T, N>::operator[](int) const’
cannot be overloaded with ‘constexpr const T& vk::ArrayWrapper1D<T, N>::operator[](int) const’
  544 |     VULKAN_HPP_CONSTEXPR T & operator[](int index) VULKAN_HPP_NOEXCEPT


oddhack:
@xantares this repository just collects files for publishing. For fixes to vulkan.hpp, you'll need to follow instructions in the README and file your issue over in https://github.com/KhronosGroup/Vulkan-Hpp . Since vulkan.hpp is a generated file, they can't accept PRs which directly modify it - changes to the generator which creates vulkan.hpp will be required. So I suggest filing an issue in that repository describing the problem, and proposed solution. Please close this PR once you've done that.

xantares:
ok

PR #160, opened 2020-11-15
pNext pointers not marked as optional in vk.xml

antlilja:
While generating language bindings from vk.xml I noticed that the pNext member in structs is not marked as optional, but from what i understand pNext should be allowed to be null. 

Example from khronos.org: "*pNext is NULL or a pointer to the next structure in a structure chain.*"
(https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkBaseInStructure.html)

Maybe I'm misinterpreting what optional means in this context.


dgkoch:
See Readme.md. XML changes should be proposed at https://github.com/KhronosGroup/Vulkan-Docs

antlilja:
Ok thank you

PR #184, opened 2021-02-24
Add support for QNX Screen platform to ICD header.

mgorchak-blackberry:
Hi,

This PR adds support for QNX Screen platform to the ICD header.

Not sure about patch for the genvk.py script, according to docs Vulkan-Docs should have the reference one, but it appears that this repository has more recent and updated genvk.py script. Same patch was already merged into the Vulkan-Docs.

Thanks!

### DIFF
--- a/include/vulkan/vk_icd.h
+++ b/include/vulkan/vk_icd.h
@@ -122,6 +122,7 @@ typedef enum {
     VK_ICD_WSI_PLATFORM_DIRECTFB,
     VK_ICD_WSI_PLATFORM_VI,
     VK_ICD_WSI_PLATFORM_GGP,
+    VK_ICD_WSI_PLATFORM_SCREEN,
 } VkIcdWsiPlatform;
 
 typedef struct {
@@ -232,5 +233,13 @@ typedef struct {
     void *window;
 } VkIcdSurfaceVi;
 #endif // VK_USE_PLATFORM_VI_NN
+
+#ifdef VK_USE_PLATFORM_SCREEN_QNX
+typedef struct {
+    VkIcdSurfaceBase base;
+    struct _screen_context *context;
+    struct _screen_window *window;
+} VkIcdSurfaceScreen;
+#endif  // VK_USE_PLATFORM_SCREEN_QNX
 
 #endif  // VKICD_H


lenny-lunarg:
Are you looking to get support for this into the [Khronos  loader](https://github.com/KhronosGroup/Vulkan-Loader)? That interface in `vk_icd.h` isn't going to do anything unless the loader can use it. The same thing goes for instance extensions in general. I don't have any plans to add loader support at the moment, but that's a small enough change that it wouldn't be a big issue. What bothers me more about adding loader support is that I have no idea how to test it because I've never touched this platform. If the QNX platform is set up so that the loader is irrelevant, let me know.

Also, one more note is that this repo updates the included Vulkan header with each patch release. We'll want to hold off merging this until we have that next patch regardless of any other considerations..

As for the actual changes, they look correct to me, but I'd normally test them by building a loader against them.

mgorchak-blackberry:
> Are you looking to get support for this into the Khronos loader? That interface in vk_icd.h isn't going to do anything unless the loader can use it.

Yes, I already did a truncated port of it some time ago, it's part of QNX future distributions. But I would like to do a fresh proper port now.

vk_icd.h also is mandatory to get Mesa3D's WSI code compiled for AMD Radeon and Intel Vulkan drivers.

> What bothers me more about adding loader support is that I have no idea how to test it because I've never touched this platform.

Current plan is to have original Vulkan ICD Loader ported to QNX to let hardware vendors test Vulkan drivers for QNX, so we get them unblocked. Since it will be the part of QNX distribution, it will be maintained in a good shape.

The difference of shipped with QNX ICD Loader only in code where it reads configuration for libraries to be loaded. It will not be a JSON file, just graphics system configuration file. But as I mentioned I will maintain both versions, default is for reference and our internal one for the shipment.

> Also, one more note is that this repo updates the included Vulkan header with each patch release. We'll want to hold off merging this until we have that next patch regardless of any other considerations..

Sure. Recent QNX support was just merged into the Vulkan-Docs, so we have to wait until it will be publicly available anyway.

> As for the actual changes, they look correct to me, but I'd normally test them by building a loader against them.

I understand. I have to go through the legal people to get a permission to post patches/PR for ICD loader, it could take about a 2-3 weeks, but sometimes less.

oddhack:
> Not sure about patch for the genvk.py script, according to docs Vulkan-Docs should have the reference one, but it appears that this repository has more recent and updated genvk.py script. Same patch was already merged into the Vulkan-Docs.

genvk.py in this repository is copied from the Vulkan-Docs repository when a spec update is done, and the next spec update will be done on Monday.

mgorchak-blackberry:
> genvk.py in this repository is copied from the Vulkan-Docs repository when a spec update is done, and the next spec update will be done on Monday.

Good, I removed it from PR.

mgorchak-blackberry:
Are there any pending showstoppers to make it moving?

Thanks!

lenny-lunarg:
I see only one problem with this, which is that github claims it can't be rebased safely. I went and rebased this locally on my machine and it worked without trouble, but it showed one extra change compared to the diff on the github PR, which was to remove the line in genvk.py:

```python
[ 'vulkan_screen.h',      [ 'VK_QNX_screen_surface'       ], commonSuppressExtensions ],
```

The diff on this PR looks great to me, but I don't know why the line above was being removed when I rebased (removing that line looks fishy to me). If you can confirm which is correct, I should be able to get it merged.

mgorchak-blackberry:
> The diff on this PR looks great to me, but I don't know why the line above was being removed when I rebased (removing that line looks fishy to me). If you can confirm which is correct, I should be able to get it merged.

Jon mentioned above, that this file will be will be overwritten using Vulkan-Docs as the source for it: https://github.com/KhronosGroup/Vulkan-Headers/pull/184#issuecomment-785526641

lenny-lunarg:
Well at this point it wouldn't get overwritten until the 1.2.172 update, since the 1.2.171 update happened yesterday. I don't know when that's happening but more often than not it's weekly updates. I'd feel better if we just got rid of the change to genvk.py (which I can do manually if need be).

mgorchak-blackberry:
> I'd feel better if we just got rid of the change to genvk.py (which I can do manually if need be).

Sure! No objections from my side.

PR #193, opened 2021-04-13
Update for Vulkan-Docs 1.2.175

oddhack:
vulkan.hpp should be getting formatted by clang-format now.


Friz64:
I think this repo is missing the `include/vk_video` directory.

Issue #214, opened 2021-07-09
Paths to sibling headers in vulkan_core.h


TheMostDiligent:
This [recent change](https://github.com/KhronosGroup/Vulkan-Headers/commit/a8964617af6b104d0341874e9b4776d82f80f9af#diff-e222ae95c2b0d5082b94d6086fb1c24da18ee31384c1a39840df3b9152023ee6R23) made all headers included from `vulkan_core.h` relative to the parent directory. This is a bit strange as all headers are located in the same folder as `vulkan_core.h`. This also caused some minor build breakages in our project.

TheMostDiligent:
In particular, `vulkan_core.h` header may not now be included as is without also adding parent directory to include search paths. This does not appear to be a correct way as internal includes should use correct relative paths.
It is also unclear what benefit using `vulkan/` gives.

PR #215, opened 2021-07-09
vulkan_core.h: fixed path to vk_platform.h (fixed https://github.com/KhronosGroup/Vulkan-Headers/issues/214)

TheMostDiligent:


oddhack:
Files in this repository are overwritten with every spec update. The headers originate from vk.xml in the Vulkan-Docs repository, which see https://github.com/KhronosGroup/Vulkan-Docs/pull/1538 for the PR that I believe caused the change here. I am closing this MR and moving the issue to Vulkan-Docs.

PR #225, opened 2021-08-12
Bump header version to 188 to match spec update

oddhack:
@KarenGhavam-lunarG someone pointed out in https://github.com/KhronosGroup/Vulkan-Docs/issues/1603 that I'd not updated the spec version number to 188 in vk.xml. I've already made this change in the Vulkan-Docs tree and updated the v1.2.188 tag there, but I'm not sure what impact it would have to your work to commit this change in the Headers repo, so am assigning this to you to decide.

If you accept the PR, then the v1.2.188 tag (https://github.com/KhronosGroup/Vulkan-Headers/releases/tag/v1.2.188) will need to be deleted and re-created to point to the commit resulting from accepting the PR.

### DIFF
--- a/include/vulkan/vulkan.hpp
+++ b/include/vulkan/vulkan.hpp
@@ -115,7 +115,7 @@ extern "C" __declspec( dllimport ) FARPROC __stdcall GetProcAddress( HINSTANCE h
 #  include <span>
 #endif
 
-static_assert( VK_HEADER_VERSION == 187, "Wrong VK_HEADER_VERSION!" );
+static_assert( VK_HEADER_VERSION == 188, "Wrong VK_HEADER_VERSION!" );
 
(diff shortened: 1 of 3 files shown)


mikes-lunarg (approved):
This change is fine to merge.  Moving the v1.2.188 tag, however, will cause CI breakage and churn in the downstream repos.  My preference would be to leave the existing tag alone and re-sync on the 189 header.  If you feel strongly that the tag should still be moved, I can work through the downstream issues. It just doesn't seem worth the effort considering we will get a fix in the next weekly update.

oddhack:
I agree it's less important to fix the inconsistency than to avoid downstream breakage. So I think this will just get closed after the 1.2.189 update without merging it, unless some emergency need arises in the next couple of days. Thanks for the quick feedback.

oddhack:
1.2.189 spec update pushed, closing.

PR #250, opened 2021-12-21
GN: Support Wayland platform

Fahien:
Following the approach taken with XCB, add support for Wayland based on user-defined variable.


null77:
LGTM

@mikes-lunarg can you let us know the process for landing this CL?

mikes-lunarg (approved):

PR #257, opened 2022-01-27
Fixed 'missing braces around initializer' warnings in vulkan_enums.hpp

sugoi1:


sugoi1:
@mikes-lunarg PTAL. Thanks

mikes-lunarg:
This file is automatically generated by the upstream project [KhronosGroup/Vulkan-Hpp](https://github.com/KhronosGroup/Vulkan-Hpp). Please file an issue in that repo instead. Once it is fixed upstream, it will propagate to this repo as part of the normal header release process.

PR #267, opened 2022-03-01
GN: Wayland include directories

Fahien:
Improve support for environments without Wayland in their system include
directories, by adding the `vulkan_wayland_include_dir` variable which can
be defined in `build_overrides`.


stonesthrow:
Are there some reviewers we can invite? This change is blocking, is there anything needed to move it along? - Thanks.

stonesthrow:
Are there some reviewers we can invite? This change is blocking, is there anything needed to move it along? - Thanks.

mikes-lunarg (approved):
This looks harmless, I will go ahead an merge to unblock folks

PR #274, opened 2022-03-29
GN: Wayland include directories

Fahien:
Wayland headers can be found in multiple directories.
Here we use the array `vulkan_wayland_include_dirs` to specify those include directories.

This is an improvement of https://github.com/KhronosGroup/Vulkan-Headers/pull/267.
Also related to [this chromium CL](https://chromium-review.googlesource.com/c/chromium/src/+/3455146/comments/c1dba934_3a7f5e43).
@null77 @mikes-lunarg

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -27,8 +27,8 @@ config("vulkan_headers_config") {
   }
   if (defined(vulkan_use_wayland) && vulkan_use_wayland) {
     defines += [ "VK_USE_PLATFORM_WAYLAND_KHR" ]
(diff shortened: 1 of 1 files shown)


Fahien:
As @stonesthrow mentioned in the previous PR, this change is still blocking, but fortunately trivial enough for a quick review.

stonesthrow:
Thank you for expediting this.

stonesthrow:
Needs reviewers

mikes-lunarg (approved):

PR #280, opened 2022-05-13
Fix build for downstream VulkanSC uses

dgkoch:
- video headers don't exist on this branch (and are not supported for Vulkan SC)
- loader interface files need to use vulkansc include instead of vulkan
- update documentation


dgkoch:
@oddhack can you merge this change for the `sc_main` branch?

oddhack:
> @oddhack can you merge this change for the `sc_main` branch?

Sure. It feels like most of these changes could go into main branch as well but not a big deal.

dgkoch:
> It feels like most of these changes could go into main branch as well but not a big deal.

Yeah - potentially in the future if we try to align everything out of the same branches.

PR #294, opened 2022-07-28
Drop wayland-client.h include

jengelh:
With https://github.com/KhronosGroup/Vulkan-Tools/pull/670 resolved, wayland-client.h can be removed from Headers. This has been cooking in openSUSE for a while, and other vulkan infrastructure (vulkan-{tools,validationlayers}, glslang, spirv-tools, shaderc) still build fine.


oddhack:
It's not possible to tell what has actually changed here, because the PR is targeting 'master' branch. That is no longer up to date and so there are a huge number of diffs - if you retarget to the current default branch, 'main', that will probably be resolved. But in any event, this would largely be reverted the next time I do a spec update, because this repository is just a staging area. Most of the content including the generator scripts and the headers comes from the Vulkan-Docs repository.

I'm guessing the actual relevant changes needed in Vulkan-Docs are simple. wayland-client.h only appears in documentation in appendices/boilerplate.txt, include/vulkan/vulkan.h, a couple of places in tests/, and a couple of places in xml/vk.xml . If you will resubmit the PR in that repo with just the relevant changes we can progress it, but we can't merge this.

PR #297, opened 2022-08-05
Fix the readme for the fixed conventions.py location

ewerness-nv:
Noticed this file had moved but hadn't been reflected in the readme


oddhack:
LGTM, thanks. The wildcard at the end of the section hopefully covers such changes, as well.

PR #207, opened 2021-06-14
Export CMake Targets File, v2

rbsheth:
Continuation of #112 #110 #89 

This should export useful CMake targets but avoid changes in any downstream projects, since changes are gated on `VULKAN_HEADERS_EXPORT_CMAKE_TARGETS` now.

PR #298, opened 2022-08-06
Adding cmake target export and install

Honeybunch:
I saw that this CMake project had no target export & install setup so I figured I could add it quickly here. 

Allows build systems like [vcpkg](https://github.com/microsoft/vcpkg) to more easily consume the cmake build. I tested this by locally applying this patch to the vcpkg port of the vulkan headers and then consuming the target in another project. 

Downstream projects can consume an installed version of the library by locating and linking with just
```
find_package(VulkanHeaders CONFIG REQUIRED)
target_link_libraries(main PRIVATE Vulkan-Headers Vulkan-Registry)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,12 +38,16 @@ endif()
 
 # define exported targets for nested project builds to consume
 add_library(Vulkan-Headers INTERFACE)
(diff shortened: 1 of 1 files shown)


Honeybunch:
I should have seen that there is a more thorough PR already open to address this: https://github.com/KhronosGroup/Vulkan-Headers/pull/207

PR #300, opened 2022-08-25
Add Bazel support

garymm:
These are pretty trivial, so I don't imagine there will be any maintenance burden, and it's helpful for consumers to not have to write this on their own.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ __pycache__
 build
 .vscode/
 **/.*.swp
+/bazel-*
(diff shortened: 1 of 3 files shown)


juan-lunarg:
We already support CMake and Chromium builds which are tested in downstream repositories like the loader and validation layers.  While adding another build system sounds trivial it never is even for header only projects.

Also because we support CMake we also end up supporting C/C++ package managers like vcpkg and conan, which ends up supporting other build systems via build system agnostic interfaces.

PR #318, opened 2022-11-07
cmake: Export CMake Targets File

### LINKED ISSUE #157: includes target "X" which requires target "Vulkan-Headers" that is not in any export set.
Currently I am looking to use Vulkan Headers directly from the repo, but it seems when I include with `target_link_library` I get the error includes target "X" which requires target "Vulkan-Headers" that is not in any export set. Is there a reason to not include an export of the vulkan-header targets? Or is there a different suggested way to add vulkan headers as part of a target that will be exported?

juan-lunarg:
Now as part of install 2 files are created:
- VulkanHeadersConfig.cmake
- VulkanHeadersConfigVersion.cmake

This allows usage of find_package

closes #157

Also closes https://github.com/KhronosGroup/Vulkan-Headers/pull/207

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,47 +14,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ~~~
-
-# CMake project initialization ---------------------------------------------------------------------------------------------------
-# This section contains pre-project() initialization, and ends with the project() command.
-
 cmake_minimum_required(VERSION 3.10.2)
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/version.cmake)
 
 project(Vulkan-Headers LANGUAGES C VERSION ${VK_VERSION_STRING})
 message(STATUS "${PROJECT_NAME} = ${PROJECT_VERSION}")
 
-# User-interface declarations ----------------------------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


juan-lunarg:
Tested against the validation layers. See draft PR for what this PR accomplishes.
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4765

CMakeLists.txt
```diff
@@ -14,47 +14,18 @@
...
-
-# uninstall target
-if(NOT TARGET uninstall)
-    configure_file("${CMAKE_CURRENT_SOURCE_DIR}/cmake/cmake_uninstall.cmake.in"
-                   "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake"
-                   IMMEDIATE
-                   @ONLY)
-    add_custom_target(uninstall COMMAND ${CMAKE_COMMAND} -P ${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake)
-endif()
+include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/install.cmake)
```
juan-lunarg, line 31: All of the install logic is inside 1 file now.

mikes-lunarg (approved):

PR #320, opened 2022-11-08
Fix duplicate quotes when targets are included by downstream projects

### LINKED ISSUE #321: Unable to integrate newer versions of Vulkan-Headers due to spurious double quotes in cmake target
In `CMakeLists.txt` the generator rules for `Vulkan-Headers` and `Vulkan-Registry` have the value on the right hand side of BUILD_INTERFACE quoted. These values should not be quoted, since the entire string, including the quotes is inserted when generating the target.

This leads to errors like this:
```
Target "Vulkan-Headers" contains relative path in its
INTERFACE_INCLUDE_DIRECTORIES:

""/home/rharrison/workspace/dawn/third_party/vulkan-deps/vulkan-headers/src/include""
```

when attempting to use these targets in downstream projects.

Removing the quotes in the target rule fixes the issue.

zoddicus:
Fixes #321

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,11 +21,11 @@ project(Vulkan-Headers LANGUAGES C VERSION ${VK_VERSION_STRING})
 message(STATUS "${PROJECT_NAME} = ${PROJECT_VERSION}")
 
 add_library(Vulkan-Headers INTERFACE)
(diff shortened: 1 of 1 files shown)


juan-lunarg:
Thanks for the quick fix @zoddicus :+1:

zoddicus:
> Thanks for the quick fix @zoddicus +1

n/p, thanks for prompt merge

PR #323, opened 2022-11-11
cmake: Update min to 3.16.3

juan-lunarg:
Currently we support the latest 2 LTS releases.

Focal 20.04 LTS ships [CMake 3.16.3](https://packages.ubuntu.com/search?suite=focal&searchon=names&keywords=cmake)


lunarpapillo:
CMake version is not tied to the Linux support matrix in any way.  It's based on the repository's functional needs, and WG partners' needs.

BUILD.md
```diff
@@ -92,7 +92,7 @@ directories and place them in any location.
     - [2019](https://www.visualstudio.com/vs/downloads/)
   - The Community Edition of each of the above versions is sufficient, as
     well as any more capable edition.
-- [CMake 3.10.2](https://cmake.org/files/v3.10/cmake-3.10.2-win64-x64.zip) is recommended.
+- [CMake 3.16+](https://cmake.org/download/) is recommended.
```
lunarpapillo, line 95: For other repositories, we specify an exact version that matches what we use in CI (as it's the only version of CMake we actually test in our builds, and thus the only version of CMake we can guarantee will work).

CMakeLists.txt
```diff
@@ -14,11 +14,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ~~~
-cmake_minimum_required(VERSION 3.10.2)
+cmake_minimum_required(VERSION 3.16.3)
```
lunarpapillo, line 17: Why 3.16.3?  It appears from the below changes that the functionality required by this file is actually supported at 3.14.0.

lunarpapillo (changes requested):

juan-lunarg:
Agreed it's not tied to the linux support matrix.

Looking into git history I thought the rational provided was persuasive.

```
It was needed for aligning to a single cmake_minimum_version across all the Vulkan repos that LunarG maintains, to reduce the maintenance burden.

Previously there were of 6 different cmake_minimum_version specified across those repos. Seemingly benign CMakeLists pull requests could trigger difficult-to-debug behaviors across the combinations of platforms/versions/repos for which LunarG runs CI and builds the Vulkan SDK.

So what single CMake version to select? After considerable discussion of trade-offs on the Vulkan mailing list (circa May 9), we arrived at CMake 3.10.2. Contributing to the decision, that version is provided by Android Studio and Ubuntu 18.04 LTS

Certainly the change will have an impact on some. Fortunately, Kitware has done a good job packaging CMake for side-by-side installs. For example: I'm running Ubuntu 16.04 LTS, which does not provide that CMake version. Unpacking the CMake 3.10.2 tarball works fine for building the Vulkan repos on 16.04.
```

BUILD.md
lunarpapillo: FYI, CI is currently at CMake 3.19.1, but will be moving to 3.23.2.

CMakeLists.txt
juan-lunarg: True, I'd be happy settling for 3.14 I picked 3.16 based on the past precedent by [mikew-lunarg](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/1029#issuecomment-546086192)

BUILD.md
juan-lunarg: Updated BUILD.md

PR #326, opened 2022-11-18
cmake: Fix environment for cross-compile installs

tgurr:
Currently the include directory is arch-dependent. However, the location where the *.cmake files are installed
is arch-independent. This difference causes an issue with cross compile environments.

### DIFF
--- a/cmake/install.cmake
+++ b/cmake/install.cmake
@@ -26,7 +26,7 @@ install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/registry" DESTINATION ${VLK_REGIS
 
 set(export_name "VulkanHeadersConfig")
 set(namespace "Vulkan::")
-set(cmake_files_install_dir ${CMAKE_INSTALL_DATADIR}/cmake/VulkanHeaders/)
+set(cmake_files_install_dir ${CMAKE_INSTALL_LIBDIR}/cmake/VulkanHeaders/)
 
 # Set EXPORT_NAME for consistency with established names. The CMake generated ones won't work.
 set_target_properties(Vulkan-Headers PROPERTIES EXPORT_NAME "Headers")


cmake/install.cmake
```diff
@@ -26,7 +26,7 @@ install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/registry" DESTINATION ${VLK_REGIS
 
 set(export_name "VulkanHeadersConfig")
 set(namespace "Vulkan::")
-set(cmake_files_install_dir ${CMAKE_INSTALL_DATADIR}/cmake/VulkanHeaders/)
+set(cmake_files_install_dir ${CMAKE_INSTALL_LIBDIR}/cmake/VulkanHeaders/)
```
juan-lunarg, line 29:
```
LIBDIR

    object code libraries (lib or lib64)
```

LIBDIR is intended for object code libraries. This approach doesn't seem correct.

juan-lunarg:
```
DATAROOTDIR

    read-only architecture-independent data root (share)
```

DATADIR is intended for files that are architecture independent. Given this is a header only library, this approach makes the most sense.

juan-lunarg: This approach is also used by [GSL](https://github.com/microsoft/GSL) FWIW which is another popular header only library.

tgurr: Right now the headers get installed into `CMAKE_INSTALL_INCLUDEDIR` which is arch-specific (e.g. in our case ends up in under the prefix `/usr/x86_64-pc-linux-gnu` in `/usr/x86_64-pc-linux-gnu/include`) and the constructed cmake files then include the arch-dependent location in `set(_IMPORT_PREFIX "/usr/x86_64-pc-linux-gnu")` like stated in the comment above so they need to be installed into the arch-dependent location `CMAKE_INSTALL_LIBDIR` as well.

Things are different for other arch-independent stuff which are no headers like the registry, it gets installed into `CMAKE_INSTALL_DATADIR` and also referenced in the cmake files under the correct arch-independent location:
`INTERFACE_INCLUDE_DIRECTORIES "/usr/share/vulkan/registry"`.

nv-codec-headers which is also a header only library (not using cmake though) also installs into INCLUDEDIR and then references the arch-dependent location in it's pkg-config files also getting installed into LIBDIR under the prefix: https://github.com/FFmpeg/nv-codec-headers/blob/master/Makefile

`/usr/x86_64-pc-linux-gnu/lib/pkgconfig/ffnvcodec.pc`
```
prefix=/usr/x86_64-pc-linux-gnu
includedir=${prefix}/include

Name: ffnvcodec
Description: FFmpeg version of Nvidia Codec SDK headers
Version: 11.1.5.1
Cflags: -I${includedir}
```
We don't have https://github.com/microsoft/GSL packaged so I can't verifiy if it's doing the right thing or also needs to be fixed.

juan-lunarg: Let me summarize if I understand correctly.

Currently the include directory is arch-dependent. However, the location where the `*.cmake` files are installed is arch-independent. This difference causes an issue with cross compile environments.

Let me know if I summarized that correctly.

tgurr: Yes, correct.

juan-lunarg: Thanks for the PR :+1:

PR #329, opened 2022-12-03
cmake: Install cmake files to CMAKE_INSTALL_DATADIR

sarnex:
If we install the cmake files into the default ABI directory, then multi-ABI builds using vulkan-headers don't work because they look for the ABI-specific cmake files which don't exist if current ABI is not the default ABI. I hit this trying to build vulkan-loader in -m32 on Gentoo, it failed now because vulkan-loader just started trying to look for the cmake files.

We need to install in the data directory.

Similar commit in spirv-headers: https://github.com/KhronosGroup/SPIRV-Headers/commit/eae955f0525724526fa602fd126623b9d84599fa

Signed-off-by: Nick Sarnie <sarnex@gentoo.org>

### DIFF
--- a/cmake/install.cmake
+++ b/cmake/install.cmake
@@ -26,7 +26,7 @@ install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/registry" DESTINATION ${VLK_REGIS
 
 set(export_name "VulkanHeadersConfig")
 set(namespace "Vulkan::")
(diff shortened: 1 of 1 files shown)


sarnex:
I just found https://github.com/KhronosGroup/Vulkan-Headers/pull/326, so probably this won't be accepted. I'll make an issue

PR #346, opened 2022-12-22
cmake: Avoid parsing version during configuration

juan-lunarg:
During future Vulkan Header updates the version will need to updated in CMakeLists.txt

The intent is to reduce complexity in the primary CMakeLists.txt and slightly increase build performance.

Many people look at the Vulkan CMake code as reference. So it's better for the community to have a clean example.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,37 +16,7 @@
 # ~~~
 cmake_minimum_required(VERSION 3.10.2)
 
-# Written as a function to minimize variable scope
-# Only VK_VERSION_STRING will be returned to the PARENT_SCOPE
-function(vlk_get_header_version)
-    set(vulkan_core_header_file "${CMAKE_CURRENT_SOURCE_DIR}/include/vulkan/vulkan_core.h")
-    if (NOT EXISTS ${vulkan_core_header_file})
-        message(FATAL_ERROR "Couldn't find vulkan_core.h!")
-    endif()
-
-    file(READ ${vulkan_core_header_file} ver)
-
-    # Get the major/minor version
-    if (ver MATCHES "#define[ ]+VK_HEADER_VERSION_COMPLETE[ ]+VK_MAKE_API_VERSION\\([ ]*[0-9]+,[ ]*([0-9]+),[ ]*([0-9]+),[ ]*VK_HEADER_VERSION[ ]*\\)")
-        set(VK_VERSION_MAJOR "${CMAKE_MATCH_1}")
-        set(VK_VERSION_MINOR "${CMAKE_MATCH_2}")
-    else()
-        message(FATAL_ERROR "Couldn't get major/minor version")
-    endif()
-
-    # Get the patch version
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -16,37 +16,7 @@
...
-        message(FATAL_ERROR "Couldn't get the patch version")
-    endif()
-
-    set(VK_VERSION_STRING "${VK_VERSION_MAJOR}.${VK_VERSION_MINOR}.${VK_HEADER_VERSION}" PARENT_SCOPE)
-endfunction()
-vlk_get_header_version()
-
-project(Vulkan-Headers LANGUAGES C VERSION ${VK_VERSION_STRING})
-message(STATUS "${PROJECT_NAME} = ${PROJECT_VERSION}")
+project(Vulkan-Headers LANGUAGES C VERSION 1.3.238)
```
juan-lunarg, line 19: @oddhack what do you think of this PR? It will require a minor modification to the CMake on new releases.

For reference this version number is important for dependent repos:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/cb16c644dee18b4fb124d45f4efcea64837f728b/layers/CMakeLists.txt#L293

tests/CMakeLists.txt
```diff
@@ -14,6 +14,34 @@
...
+if (ver MATCHES "#define[ ]+VK_HEADER_VERSION[ ]+([0-9]+)")
+    set(VK_HEADER_VERSION "${CMAKE_MATCH_1}")
+else()
+    message(FATAL_ERROR "Couldn't get the patch version")
+endif()
+
+set(VK_VERSION_STRING "${VK_VERSION_MAJOR}.${VK_VERSION_MINOR}.${VK_HEADER_VERSION}")
+
+if (NOT ${VK_VERSION_STRING} STREQUAL ${PROJECT_VERSION})
+    message(FATAL_ERROR "Project version is ${PROJECT_VERSION} but should be ${VK_VERSION_STRING}!")
```
juan-lunarg, line 42: Since CI enables testing, this will catch if the version hasn't been properly updated.

juan-lunarg: The error message will look like this:
```
[cmake] CMake Error at tests/CMakeLists.txt:42 (message):
[cmake]   Project version is 1.2.238 but should be 1.3.238!
```

CMakeLists.txt
oddhack: I would rather it extract the version number from the XML or header that's packaged in the repo, if you can express something like

> grep -E '^#define VK_HEADER_VERSION [1-9][0-9]+$' | awk '{print $3}'

where the version number is expected. Every time something gets updated independently of the upstream source, there's a chance that it won't get updated correctly.

oddhack: If you're saying you want it hardwired in the file so if it's used outside the repo inconsistencies will be caught, perhaps a CI job triggered on commits to main could take care of this. My only role in the repo is updating the artifacts from Vulkan-Docs and Vulkan-Hpp and I'm loathe to take on responsibility for part of a build system I'm not involved with and don't understand.

PR #350, opened 2023-01-04
Deprecate vk_sdk_platform.h

### LINKED ISSUE #316: vk_sdk_platform.h should be removed
This file contains some very minor platform / compiler macro definitions that IMO don't belong in Vulkan-Headers at all. None of the macros in this file are used by any of the other headers in Vulkan-Headers.  Applications that need them would be better served by defining what they need in their own cross platform code.

However, this header is included by several repositories on github, including some not part of KhronosGroup.  There is likely other code not on github using this header.  Any effort to remove the header file would probably need to be staged carefully, and include a phase where the header is marked deprecated somehow so that people have time to fix stuff before it is removed completely.

juan-lunarg:
closes #316


charles-lunarg (commented):
Is there anything specific needed for clang?

include/vulkan/vk_sdk_platform.h
```diff
@@ -20,6 +20,20 @@
  */
 #pragma once
 
+// Allow users to suppress warnings generated by this header file by defining VK_SDK_PLATFORM_SUPRRESS_DEPRECATION_WARNING
+#ifndef VK_SDK_PLATFORM_SUPRRESS_DEPRECATION_WARNING
+
+#if defined(__GNUC__) && __GNUC__ >= 4
```
juan-lunarg, line 26: The oldest version of GCC available on godbolt was version 4.0

juan-lunarg:
> Is there anything specific needed for clang?

[The GNU check is compatible with clang 3.0](https://godbolt.org/z/b4zjxon3E)

charles-lunarg (approved):
Works for me! I would of guessed clang followed gcc here but I wanted to be sure.

juan-lunarg:
Clang does follow GCC. Clang also implements `__GNUC__` and `#warning`.

I'm confused what you mean?

charles-lunarg:
I mean that clang implements the same behavior as gcc, and there is no need to have a check for it because all versions of clang support this feature. Apologies for the confusing statements.

jeremyg-lunarg (approved):
LGTM

jeremyg-lunarg:
> I mean that clang implements the same behavior as gcc, and there is no need to have a separate check for it because all versions of clang support this feature. Apologies for the confusing statements.

Fun fact clang also sets _MSC_VER https://clang.llvm.org/docs/UsersManual.html#microsoft-extensions

juan-lunarg:
> Fun fact clang also sets _MSC_VER https://clang.llvm.org/docs/UsersManual.html#microsoft-extensions

Looks like enabling the MSVC functionality requires usage of `-fms-extensions` :+1:

PR #352, opened 2023-01-09
cmake: Remove Vulkan::Registry

### LINKED ISSUE #351: Setting target_include_directories with generator expression breaks configuration of layers
**Describe the bug**

`Vulkan-Headers` and `Vulkan-ValidationLayers` are included in my project as submodules via `add_subdirectory`.
During CMake configuration step of `Vulkan-ValidationLayers` I get an error at [line](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/4050be7c1960ae0aec00b421bb47be8d67cabe63/layers/CMakeLists.txt#L25) where `VulkanRegistry_DIR` get `$<BUILD_INTERFACE:/home/anatoliy/Projects/sah_kd_tree/external/Vulkan-Headers/registry>`. As you can see there is unevaluated generator expression, that comes directly from the [line](https://github.com/KhronosGroup/Vulkan-Headers/blob/24115c70bea4bb91e7bd12ef093fbdf6aa9a5222/CMakeLists.txt#L56). It is expected, because *generator expression* evaluated at step when *generator works*, not at *configuration step*. Therefore `EXISTS` failed.

**Expected behavior**
Project configured.

juan-lunarg:
Vulkan::Registry behaves differently in `add_subdirectory` vs
`find_package` builds. Which can result in build failures.

Also Vulkan::Registry never made sense as a CMake target and
was very brittle in numerous ways.

Better to provide a `VULKAN_HEADERS_REGISTRY_DIRECTORY` to users
instead and remove Vulkan::Registry completely.

closes #351

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -54,6 +54,8 @@ Which, when successful, will add library target called `Vulkan::Headers` which y
 find_package(VulkanHeaders REQUIRED CONFIG)
 
 target_link_libraries(foobar PRIVATE Vulkan::Headers)
+
+message(STATUS "Vulkan Headers Registry: ${VULKAN_HEADERS_REGISTRY_DIRECTORY}")
 ```
 
 ## Repository Set-Up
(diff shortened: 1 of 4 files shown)


juan-lunarg:
Tested against VVL and it actually makes the CMake code cleaner.

charles-lunarg (approved):
Looks good assuming the commit message is changed to reflect how its being removed, not just deprecated.

juan-lunarg:
> Looks good assuming the commit message is changed to reflect how its being removed, not just deprecated.

Updated commit message / PR

PR #356, opened 2023-01-11
cmake: Removes unused CMake macros

juan-lunarg:


abouvier:
You should not remove the `check_required_components` macro. Now `find_package(VulkanHeaders REQUIRED COMPONENTS NonExistingComponent)` will not raise a fatal error.

PR #361, opened 2023-01-16
ci: Cleanup linux.yml

juan-lunarg:


.github/workflows/linux.yml
```diff
@@ -61,5 +58,8 @@ jobs:
           cmakeVersion: 3.10.2
       - name: CMake Version
         run: cmake --version
+      - name: Create build directory
+        run: mkdir build/
       - name: Configure Vulkan-Headers
-        run: mkdir build/ && cd build/ && cmake ${GITHUB_WORKSPACE}
+        working-directory: ./build
+        run: cmake ../
```
MarijnS95, line 65: I'd say to just run `cmake -Bbuild`

MarijnS95: Just like above, apparently.

MarijnS95: It's just a suggestion to simplify these lines, to match what's done above; you're only configuring a cmake project here without building it (that'd be `cmake --build path/to/generated/build/folder`).

juan-lunarg: `cmake --build ...` won't configure a project unless a it already exists. So I'm confused what you mean.

juan-lunarg:
```
juan@pop-os:~/projects/pub-headers$ ls
BUILD.gn  BUILD.md  cmake  CMakeLists.txt  CODE_OF_CONDUCT.md  CONTRIBUTING.md  include  INTEGRATION.md  LICENSE.txt  README.md  registry  tests
juan@pop-os:~/projects/pub-headers$ cmake --build build/
Error: /home/juan/projects/pub-headers/build is not a directory
```

MarijnS95: @juan-lunarg I said to use `cmake -B build`. `-B` **is not** a short-hand for `--build`!

juan-lunarg: Ah. That makes more sense. Sorry for the confusion that's my bad.

Sadly CMake `3.10.2` doesn't have `-S` / `-B` they were introduced later.

juan-lunarg: Yep `-S` and `-B` were introduced in 3.13 and we are explicitly testing the cmake_minimum advertised

MarijnS95: Ah that explains, was already wondering!

PR #364, opened 2023-01-19
docs: Remove un-neccessary detail from docs

juan-lunarg:


oddhack:
I'd like to revert (most) of this. The comments about the headers are there because periodically someone will post a PR in this repo asking for a change to headers that are sourced from another repo, and I think the versioning scheme explanation is useful (perhaps not the historical part of it, though).

juan-lunarg:
> I'd like to revert (most) of this. The comments about the headers are there because periodically someone will post a PR in this repo asking for a change to headers that are sourced from another repo, and I think the versioning scheme explanation is useful (perhaps not the historical part of it, though).

Fixed https://github.com/KhronosGroup/Vulkan-Headers/pull/368

juan-lunarg:
> The comments about the headers are there because periodically someone will post a PR

I've also made a PR template to try and help with this issue and verified it works.

https://github.com/KhronosGroup/Vulkan-Headers/commit/1a43b498b8e680d641af0e0a4b4bb47947b7a364
https://github.com/KhronosGroup/Vulkan-Headers/pull/370

PR #376, opened 2023-01-30
cmake: Trim CMakeLists.txt

juan-lunarg:


juan-lunarg:
Documentation is overkill now that testing covers the find_package support

Changing this code incorrectly will now cause CI to fail

PR #377, opened 2023-02-08
headers: Remove VK_LAYER_EXPORT

juan-lunarg:
- It hasn't been handling windows (so it's been half implemented)
- It's a suboptimal method for exporting symbols for vulkan layers since `--version-script` and `.def` files exists. Both of these methods also don't risk name mangling.
- `--version-script` also has perf benefits since you can silence exports from code you didn't intend to export

### DIFF
--- a/include/vulkan/vk_layer.h
+++ b/include/vulkan/vk_layer.h
@@ -28,14 +28,6 @@
 
 #include "vulkan_core.h"
 
-#if defined(__GNUC__) && __GNUC__ >= 4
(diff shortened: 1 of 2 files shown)


charles-lunarg (approved):
Good to remove code which really doesn't work that well.
I will say that there may be people out there who use this code, but its trivial to copy/paste this into a codebase. Plus, its likely that if they need this macro, they also need to handle the cases where this doesn't work. 

The loader uses it in one place, except that it defines it if it isn't already (which this should of been doing...) so this change is fine.

PR #379, opened 2023-02-17
Add -Wno-redundant-parens to BUILD.gn

mikes-lunarg:
The SC vk.xml merge added extra parentheses to the VK_DEFINE_NON_DISPATCHABLE_HANDLES macro in order to satisfy MISRA rule 20.7.  Make sure this doesn't break downstream builds that would otherwise warn about them.


mark-lunarg (approved):
(((LGTM))).

juan-lunarg (approved):

PR #383, opened 2023-03-08
Allow finding Python in other locations

brad0:
Although a lot of Linux OS's provide Python in /usr/bin that is not always the case and outside of Linux that is not the case.

### DIFF
--- a/registry/apiconventions.py
+++ b/registry/apiconventions.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3 -i
+#!/usr/bin/env python3 -i
 #
 # Copyright 2021-2023 The Khronos Group Inc.
 # SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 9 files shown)


juan-lunarg:
Looks like a good PR but please see the CONTRIBUTING.md
https://github.com/KhronosGroup/Vulkan-Headers/blob/main/CONTRIBUTING.md

These *.py files are imported from the Specification repository (https://github.com/KhronosGroup/Vulkan-Docs). So you'll want to make the PR there.

See the CONTRIBUTING.md to see which files are maintained here.

brad0:
> Looks like a good PR but please see the CONTRIBUTING.md https://github.com/KhronosGroup/Vulkan-Headers/blob/main/CONTRIBUTING.md
> 
> These *.py files are imported from the Specification repository (https://github.com/KhronosGroup/Vulkan-Docs). So you'll want to make the PR there.
> 
> See the CONTRIBUTING.md to see which files are maintained here.

Oh, my bad. I'll submit an equivalent patch over there. I see there are more Python scripts over there.

PR #409, opened 2023-06-28
cmake: unguard install rules

abouvier:
Disabling install rules for subprojects is wrong. For example this code will raise an error:
```
add_library(mylib STATIC ...)
target_link_libraries(mylib PUBLIC Vulkan::Headers)
install(TARGETS mylib EXPORT mylibTargets)
install(EXPORT mylibTargets ...)
```
```
$ cmake ...
...
CMake Error: install(EXPORT "mylibTargets" ...) includes target "mylib" which requires target "Vulkan-Headers" that is not in any export set.
```
Because `mylib` is a static library and all its dependencies (even interface libraries) are needed by programs that links with it.
If you don't want those files (when it's possible) you can use the keyword `EXCLUDE_FROM_ALL`:
```
add_library(mylib SHARED ...)
add_subdirectory(Vulkan-Headers EXCLUDE_FROM_ALL)
target_link_libraries(mylib PUBLIC Vulkan::Headers)
...
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,52 +43,56 @@ vlk_get_header_version()
 
 project(VULKAN_HEADERS LANGUAGES C VERSION ${VK_VERSION_STRING})
 
-add_library(Vulkan-Headers INTERFACE)
-add_library(Vulkan::Headers ALIAS Vulkan-Headers)
-target_include_directories(Vulkan-Headers INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>)
-
 if (CMAKE_VERSION VERSION_LESS "3.21")
     # https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html
     string(COMPARE EQUAL ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_SOURCE_DIR} PROJECT_IS_TOP_LEVEL)
 endif()
+
+if (NOT PROJECT_IS_TOP_LEVEL)
+    set_directory_properties(PROPERTIES EXCLUDE_FROM_ALL ON)
+endif()
+
+add_library(Vulkan-Headers INTERFACE)
+add_library(Vulkan::Headers ALIAS Vulkan-Headers)
+target_include_directories(Vulkan-Headers INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>)
 
 if (PROJECT_IS_TOP_LEVEL)
     option(BUILD_TESTS "Build the tests")
     if (BUILD_TESTS)
         add_subdirectory(tests)
     endif()
-
-    include(GNUInstallDirs)
-    include(CMakePackageConfigHelpers)
-
-    set(VLK_REGISTRY_DIR "${CMAKE_INSTALL_DATADIR}/vulkan")
-
-    install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/include/vk_video" DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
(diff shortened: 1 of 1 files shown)


juan-lunarg (changes requested):
Not all projects that call `add_subdirectory(Vulkan-Headers)` want to install Vulkan-Headers as part of their install process.

abouvier:
That's why they need to use `add_subdirectory(Vulkan-Headers EXCLUDE_FROM_ALL)`. But currently it's impossible to have these install rules executed within a subdirectory.

juan-lunarg:
1.) I'd rather have installation behind an `option` I see precedent for it in multiple popular CMake projects like the Microsoft Guideline Support Library.

https://github.com/KhronosGroup/Vulkan-Headers/pull/409
> option(GSL_INSTALL "Generate and install GSL target" ${PROJECT_IS_TOP_LEVEL})

Other project like GLFW, google tests, etc. also provide an option to enable or disable installation.

2.) I want the default for installation to be ON if the project is top level. OFF otherwise.

This plays nicely with solutions like FetchContent and CPM.

3.) Relying on `EXCLUDE_FROM_ALL`

This is an excerpt from my copy [Professional CMake](https://crascit.com/professional-cmake/)

> The optional EXCLUDE_FROM_ALL keyword is intended to control whether targets defined in the
subdirectory being added should be included in the project’s ALL target by default. Unfortunately,
for some CMake versions and project generators, it doesn’t always act as expected and can even
result in broken builds.

juan-lunarg:
Please see my PR for an alternative solution to this problem:

https://github.com/KhronosGroup/Vulkan-Headers/pull/411/files

abouvier:
OK fine!
Just for fun I modified this PR to still use `EXCLUDE_FROM_ALL` but automatically, so the final user doesn't have to specify the keyword himself in `add_subdirectory(Vulkan-Headers)`.

CMakeLists.txt
```diff
@@ -43,52 +43,56 @@ vlk_get_header_version()
...
-target_include_directories(Vulkan-Headers INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>)
-
 if (CMAKE_VERSION VERSION_LESS "3.21")
     # https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html
     string(COMPARE EQUAL ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_SOURCE_DIR} PROJECT_IS_TOP_LEVEL)
 endif()
 
+if (NOT PROJECT_IS_TOP_LEVEL)
+    set_directory_properties(PROPERTIES EXCLUDE_FROM_ALL ON)
+endif()
```
juan-lunarg, line 53: I'd rather not use this solution since it's not very idiomatic. I.E. I have never seen this approach before. I like CMake and even I find this hard to read.

> Unfortunately, for some CMake versions and project generators, it doesn’t always act as expected and can even
result in broken builds.

This still creates an `INSTALL` target on visual studio generated projects that consume this project with add_subdirectory. This `INSTALL` target installs nothing which is pretty confusing. IE this doesn't work nicely on all platforms.

juan-lunarg:
Closing in favor of https://github.com/KhronosGroup/Vulkan-Headers/pull/411

juan-lunarg:
```
add_library(mylib STATIC ...)
target_link_libraries(mylib PUBLIC Vulkan::Headers)
install(TARGETS mylib EXPORT mylibTargets)
install(EXPORT mylibTargets ...)
```

Also FWIW be careful about exposing another library interface in your library interface. It can make your library more difficult to consume for other users. It's generally not recommended to expose usage of another library in your libraries PUBLIC interface unless you absolutely have to.

Note: I don't know your exact situation I'm just offering general advice learned from painful experience.

abouvier:
It's not me, it's [VulkanMemoryAllocator](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/blob/0aa3989b8f382f185fdf646cc83a1d16fa31d6ab/src/CMakeLists.txt#L30) :p

PR #411, opened 2023-06-29
cmake: Add VULKAN_HEADERS_INSTALL option

juan-lunarg:
Allows add_subdirectory users to install vulkan-headers

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,7 +57,10 @@ if (PROJECT_IS_TOP_LEVEL)
     if (BUILD_TESTS)
         add_subdirectory(tests)
     endif()
+endif()
 
+option(VULKAN_HEADERS_INSTALL "Install Vulkan Headers" ${PROJECT_IS_TOP_LEVEL})
+if (VULKAN_HEADERS_INSTALL)
     include(GNUInstallDirs)
     include(CMakePackageConfigHelpers)
 
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -19,6 +19,9 @@ on:
     pull_request:
         branches:
             - main
+
+env:
+  CMAKE_GENERATOR: Ninja
             
 permissions:
     contents: read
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -57,7 +57,10 @@ if (PROJECT_IS_TOP_LEVEL)
     if (BUILD_TESTS)
         add_subdirectory(tests)
     endif()
+endif()
 
+option(VULKAN_HEADERS_INSTALL "Install Vulkan Headers" ${PROJECT_IS_TOP_LEVEL})
+if (VULKAN_HEADERS_INSTALL)
```
juan-lunarg, line 63: Here are a couple repos I found that make installation an option.

- https://github.com/glfw/glfw/blob/3fa2360720eeba1964df3c0ecf4b5df8648a8e52/CMakeLists.txt#L21
- https://github.com/microsoft/GSL/blob/87e21400dc695e45412fb0a512c73d8d4aaa2d5b/CMakeLists.txt#L11
- https://github.com/jbeder/yaml-cpp/blob/b8882652fcbeba4c00dec019a39da91e702e474e/CMakeLists.txt#L26
- https://github.com/google/googletest/blob/687c589949eaf8734484491a6882a7fc56aebc12/CMakeLists.txt#L20

juan-lunarg: Wouldn't this hide `VULKAN_HEADERS_INSTALL` from you preventing your use case entirely?

juan-lunarg: From the documentation of `cmake_dependent_option`

>  Makes <option> available to the user if the [semicolon-separated list](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#cmake-language-lists) of conditions in <depends> are all true. Otherwise, a local variable named <option> is set to <force>.

Wouldn't this make `VULKAN_HEADERS_INSTALL` impossible to be set to `ON` by an `add_subdirectory` user? Testing it locally that seems to be the case. But maybe I missed something.

abouvier: Damn, I think you're right. `cmake_dependent_option` destroys any previously defined variable.

juan-lunarg:
@abouvier you can now install Vulkan-Headers with add_subdirectory. You just need to set `VULKAN_HEADERS_INSTALL` to `ON`. Here is some example code.

```cmake
set(VULKAN_HEADERS_INSTALL ON)
```

PR #416, opened 2023-07-05
cmake: Remove usage of configure_package_config_file

juan-lunarg:
Fixes issue shown here:
https://github.com/KhronosGroup/Vulkan-Headers/pull/415

This also makes Vulkan-Headers more consistent with SPIRV-Headers


juan-lunarg:
There is still a config file. We actively test find_package support as part of CI.

juan-lunarg:
`configure_package_config_file` isn't needed to generate a config file.

CMakeLists.txt
```diff
@@ -71,27 +71,14 @@ if (VULKAN_HEADERS_INSTALL)
...
-        NO_SET_AND_CHECK_MACRO
-        NO_CHECK_REQUIRED_COMPONENTS_MACRO
-    )
-
-    set(config_version "${CMAKE_CURRENT_BINARY_DIR}/VulkanHeadersConfigVersion.cmake")
+    set(cmake_install_dir "${CMAKE_INSTALL_DATADIR}/cmake/VulkanHeaders")
 
-    write_basic_package_version_file(${config_version} COMPATIBILITY SameMajorVersion ARCH_INDEPENDENT)
+    install(TARGETS Vulkan-Headers EXPORT VulkanHeadersConfig INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+    install(EXPORT VulkanHeadersConfig NAMESPACE "Vulkan::" DESTINATION "${cmake_install_dir}")
```
juan-lunarg, line 79: These 2 lines are enough to automatically generate a config file named `VulkanHeadersConfig`

juan-lunarg:
```
cmake --install build --prefix /home/runner/work/Vulkan-Headers/Vulkan-Headers/build/install
-- Installing: /home/runner/work/Vulkan-Headers/Vulkan-Headers/build/install/share/cmake/VulkanHeaders/VulkanHeadersConfig.cmake
```

abouvier:
Yes there is a config file but only after installation. The goal was to have a config file in the the build tree, which can be used by subprojects linking with VulkanHeaders:
```
set(VulkanHeaders_DIR "${CMAKE_CURRENT_BINARY_DIR}/Vulkan-Headers") # previous location of the VulkanHeadersConfig generated by configure_package_config_file
add_subdirectory(ProjectUsingVulkanHeaders) # will invoke find_package(VulkanHeaders)
```

juan-lunarg:
> The goal was to have a config file in the the build tree, which can be used by subprojects linking with VulkanHeaders:

The CMake code you are showing is full of various issues and is not a supported use case.

juan-lunarg:
It's clear your issue stems from the CMake code in vulkan memory allocator. I've made a PR to address the various CMake issues in VulkanMemoryAllocator.

abouvier:
> The CMake code you are showing is full of various issues and is not a supported use case.
> 
> You are mixing add_subdirectory and find_package of the same dependency in order to work around another dependency.


It's the equivalent of this code but for old CMake versions:
```
include(FetchContent)
FetchContent_Declare(
    VulkanHeaders
    SOURCE_DIR "${CMAKE_CURRENT_LIST_DIR}/Vulkan-Headers"
    OVERRIDE_FIND_PACKAGE # introduced in CMake 3.24
)
FetchContent_Declare(
    ProjectUsingVulkanHeaders
    SOURCE_DIR "${CMAKE_CURRENT_LIST_DIR}/ProjectUsingVulkanHeaders"
)
FetchContent_MakeAvailable(VulkanHeaders ProjectUsingVulkanHeaders)
```

juan-lunarg:
My proposed changes to VMA (https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/pull/349) will allow the following use cases to work without any issue.

```
add_subdirectory(VulkanMemoryAllocator)
```

```
find_package(Vulkan-Headers)
add_subdirectory(VulkanMemoryAllocator)
```

```
find_package(VulkanMemoryAllocator)
```

```
find_package(Vulkan-Headers)
find_package(VulkanMemoryAllocator)
```

abouvier:
> My proposed changes to VMA ([GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator#349](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/pull/349)) will allow the following use cases to work without any issue.

Of course, there is no more `find_package(Vulkan)` and/or linking with `Vulkan::Headers` by default in your PR :p
But the problem still persists for projects linking with `Vulkan::Headers`. There is no way to hijack their `find_package(VulkanHeaders)` to use a local/custom `Vulkan::Headers` added via `add_subdirectory(Vulkan-Headers)`.

juan-lunarg:
> There is no way to hijack their find_package(VulkanHeaders) to use a local/custom Vulkan::Headers added via add_subdirectory(Vulkan-Headers).

I asked this question to the CMake developers directly:
https://discourse.cmake.org/t/idiomatic-way-to-handle-packages-and-add-subdirectory/8400

This isn't the responsibility of VulkanHeaders.

abouvier:
Look what I found in your _Professional CMake_ book written by a CMake co-maintainer 😱
Screenshot_20230708_214103

juan-lunarg:
Yes I was already aware of that. I looked at the config file generated by this approach and saw no downside since it had parity with the existing solution.

For more complicated projects it may be needed. Not for vulkan headers though.

PR #418, opened 2023-07-05
cmake: Remove VULKAN_HEADERS_INSTALL

juan-lunarg:
VULKAN_HEADERS_INSTALL was added since it was believed there was a valid use case for it.

After looking into the use case that provoked this change there is no reason to keep VULKAN_HEADERS_INSTALL as an option.

Here is the use case we do NOT want to support:
https://github.com/KhronosGroup/Vulkan-Headers/pull/416#issuecomment-1622318949

Fundamentally this problem is caused add_subdirectory/find_package not being able to work together flawlessly.

Which isn't the responsibility of Vulkan-Headers to fix.

It's the responsibility of projects that consume Vulkan-Headers to account for either method.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,19 +57,14 @@ if (PROJECT_IS_TOP_LEVEL)
     if (BUILD_TESTS)
         add_subdirectory(tests)
     endif()
-endif()
 
-option(VULKAN_HEADERS_INSTALL "Install Vulkan Headers" ${PROJECT_IS_TOP_LEVEL})
-if (VULKAN_HEADERS_INSTALL)
     include(GNUInstallDirs)
     include(CMakePackageConfigHelpers)
-
-    set(VLK_REGISTRY_DIR "${CMAKE_INSTALL_DATADIR}/vulkan")
 
     install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/include/vk_video" DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
     install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/include/vulkan" DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
     # Preserve source permissions https://github.com/KhronosGroup/Vulkan-Headers/issues/336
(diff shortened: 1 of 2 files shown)


abouvier:
Don't break everything lol, `VULKAN_HEADERS_INSTALL` was added for a different problem.
For example this `CMakeLists.txt` won't work without this option:
```
cmake_minimum_required(VERSION 3.15)

project(lol)

add_library(lol STATIC main.c)
target_link_libraries(lol PUBLIC Vulkan::Headers)

install(TARGETS lol EXPORT lolTargets)
install(EXPORT lolTargets NAMESPACE lol:: DESTINATION lib/cmake/lol)
```
```
$ cmake -S . -B build
-- The C compiler identification is GNU 13.1.1
-- The CXX compiler identification is GNU 13.1.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/lib/ccache/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/lib/ccache/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done (0.2s)
CMake Error: install(EXPORT "lolTargets" ...) includes target "lol" which requires target "Vulkan-Headers" that is not in any export set.
-- Generating done (0.0s)
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

juan-lunarg:
That could potentially result in 2 different versions of `vulkan-headers` being installed by a package manager. 

One from Vulkan-Headers and another from `lol`

Which would cause one install to override another potentially misconfiguration the machine.

abouvier:
Yes, but it's not your problem :p
If the creator of this library doesn't want to use `find_package(VulkanHeaders)` with a system Vulkan-Headers package, it's probably because there is none installed on his machine in the first place.
And maybe this library is itself used as a subproject:
```
add_subdirectory(lol EXCLUDE_FROM_ALL)
```
So the Vulkan-Headers files won't even be installed along with the top program.

PR #424, opened 2023-07-17
ci: Add Windows to CI

juan-lunarg:


juan-lunarg:
CI passed fine on my fork.
https://github.com/juan-lunarg/Vulkan-Headers/actions/runs/5579359004

Not sure why but actions is really slow today.

PR #453, opened 2023-10-23
gn: Fix validation layer build

juan-lunarg:
closes KhronosGroup/Vulkan-ValidationLayers/issues/6780

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -27,8 +27,14 @@ config("vulkan_headers_config") {
   if (is_fuchsia) {
     defines += [ "VK_USE_PLATFORM_FUCHSIA" ]
   }
-  if (is_mac) {
+  if (is_apple) {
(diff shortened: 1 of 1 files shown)


mikes-lunarg (approved):

juan-lunarg:
Today I learned you can close issues from other repos like I showed above. Neat 👍🏾

oddhack:
Convenient but may also have some pitfalls. I've noticed with the gitlab vulkan and vulkansc repos, which share a commit history for the most part, that references in commit comments to issue/MR numbers can cause odd activity when it reaches the other repo, which may have entirely unrelated issues/MRs with the same numbers. I feel like this is an oversight in automatic push/pull synchronization, but rewriting the commit comment in the process has its own pitfalls. So far I don't *think* it's resulted in accidentally closing something that should not have been, at least. The most visible result has been issues/MRs getting activity notifications when they were closed ages ago.

PR #471, opened 2024-02-13
cmake: Allow external control of whether to test or install

dneto0:
This makes the project more composable: It can be built and tested as part of a larger set of projects, from source.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,13 +47,15 @@ if (CMAKE_VERSION VERSION_LESS "3.21")
     string(COMPARE EQUAL ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_SOURCE_DIR} PROJECT_IS_TOP_LEVEL)
 endif()
 
-if (PROJECT_IS_TOP_LEVEL)
-    option(BUILD_TESTS "Build the tests")
-    if (BUILD_TESTS)
-        enable_testing()
-        add_subdirectory(tests)
-    endif()
+option(VULKAN_HEADERS_ENABLE_TESTS "Test Vulkan-Headers" ${PROJECT_IS_TOP_LEVEL})
+option(VULKAN_HEADERS_ENABLE_INSTALL "Install Vulkan-Headers" ${PROJECT_IS_TOP_LEVEL})
 
+if (VULKAN_HEADERS_ENABLE_TESTS)
+    enable_testing() # This is only effective in the top level CMakeLists.txt file.
+    add_subdirectory(tests)
+endif()
+
+if (VULKAN_HEADERS_ENABLE_INSTALL)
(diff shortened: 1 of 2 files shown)


charles-lunarg (changes requested):
Because this changes the name of BUILD_TESTS, it breaks anyone's local build but also breaks the github actions CI job. 

https://github.com/KhronosGroup/Vulkan-Headers/blob/5ac36269f50381bdd92a5e1973d8eb041771e59e/.github/workflows/ci.yml#L32

While I'm not sure there is a policy around the naming, the `BUILD_TESTS` was consistent across other Vulkan repositories. Having a per-repo variable allows tighter control of which repo's are doing what when vendored, but it is not common for vendored repo's to want to test their dependencies. 

Put another way, if you want to change the variable name, at least change all occurrences of it in the codebase.

CMakeLists.txt
```diff
@@ -47,13 +47,16 @@ if (CMAKE_VERSION VERSION_LESS "3.21")
...
+option(VULKAN_HEADERS_ENABLE_TESTS "Test Vulkan-Headers" ${PROJECT_IS_TOP_LEVEL})
+option(VULKAN_HEADERS_ENABLE_INSTALL "Install Vulkan-Headers" ${PROJECT_IS_TOP_LEVEL})
+
+if (VULKAN_HEADERS_ENABLE_TESTS)
+    enable_testing() # This is only effective in the top level CMakeLists.txt file.
+    add_subdirectory(tests)
+endif()
 
+if (VULKAN_HEADERS_ENABLE_INSTALL)
+	message("enabling install")
```
juan-kitware, line 59: By default this will be directed to stderr. I'd just remove the message. Or change the mode to DEBUG instead.

CMakeLists.txt
```diff
@@ -47,14 +47,16 @@ if (CMAKE_VERSION VERSION_LESS "3.21")
...
+option(VULKAN_HEADERS_ENABLE_INSTALL "Install Vulkan-Headers" ${PROJECT_IS_TOP_LEVEL})
+
+if (VULKAN_HEADERS_ENABLE_TESTS)
+    enable_testing() # This is only effective in the top level CMakeLists.txt file.
+    add_subdirectory(tests)
+endif()
 
-    include(GNUInstallDirs)
+if (VULKAN_HEADERS_ENABLE_INSTALL)
+	include(GNUInstallDirs)
```
juan-kitware, line 59: indentation issue

charles-lunarg (approved):
I went ahead and applied the name change into ci.yml. VULKAN_HEADERS_ENABLE_TESTS is a mouthful but it is very explicit and wont be enabled by accident.

juan-kitware (approved):
I'd rebase and squash to make this into 1 commit. Other than that LGTM

PR #484, opened 2024-06-09
Add a new `IMPORTED` target `Vulkan::VulkanHppModule` which contains `vulkan.cppm`

sharadhr:
Added a new `IMPORTED` target and CMake library, `Vulkan::VulkanHppModule` that consumers can link against, to transparently use `import vulkan_hpp;`
- Avoids users having to [set the library up themselves](https://github.com/KhronosGroup/Vulkan-Hpp?tab=readme-ov-file#usage-with-cmake)
- Code mostly derived from [CMake's own tests](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Tests/RunCMake/CXXModules/examples/export-with-headers-install/CMakeLists.txt)
- Guarded this target behind a version test requiring at least CMake version 3.28

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,8 +48,9 @@ if (MSVC AND (MSVC_VERSION GREATER_EQUAL "1934") OR
     set(COMPILER_SUPPORTS_CXX_MODULES TRUE)
 endif()
 
+option(VULKAN_HEADERS_ENABLE_MODULE "Enables building of the Vulkan C++ module. Default is true if supported by the CMake version and compilers" ${COMPILER_SUPPORTS_CXX_MODULES})
 
-if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.28" AND COMPILER_SUPPORTS_CXX_MODULES)
+if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.28" AND VULKAN_HEADERS_ENABLE_MODULE)
     add_library(Vulkan-Module)
     add_library(Vulkan::VulkanHppModule ALIAS Vulkan-Module)
     target_sources(Vulkan-Module
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 # ~~~
-cmake_minimum_required(VERSION 3.15...3.25)
+cmake_minimum_required(VERSION 3.15)
 
 # NOTE: Parsing the version like this is suboptimal but neccessary due to our release process:
 # https://github.com/KhronosGroup/Vulkan-Headers/pull/346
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -38,14 +38,23 @@ vlk_get_header_version()
...
-    # https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html
-    string(COMPARE EQUAL ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_SOURCE_DIR} PROJECT_IS_TOP_LEVEL)
-endif()
+add_library(Vulkan-Module)
+add_library(Vulkan::VulkanHppModule ALIAS Vulkan-Module)
+target_sources(Vulkan-Module
+    PUBLIC
+        FILE_SET module CXX_MODULES
+            BASE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}/include"
+            FILES "${CMAKE_CURRENT_SOURCE_DIR}/include/vulkan.cppm"
```
charles-lunarg, line 55: Use of CMake features this new should be guarded by if-checks, like the one seen on line 45 for CMake 3.21.

charles-lunarg (changes requested):
While I appreciate the improvements to Vulkan-Hpp that allow module usage, it cannot create errors for existing users by updating the CMake version unilaterally. 3.28 is extremely new and would punish anyone without the bleeding edge.

CMakeLists.txt
sharadhr: Thanks for the review. Would 3.23 be more acceptable? This is the earliest version that introduces `FILE_SET`, which would also allow dropping the test in line 45.

sharadhr: Cheers. I've guarded this behind a similar `if` statement, and restored line 45 itself.

charles-lunarg (commented):
Changes look appropriate for what is being done - I would suggest squashing the commits so that the intermediate steps aren't in the git history. Same with the github PR description being out of date with the current set of changes.

sharadhr:
Squashed commits as requested. I've also made a few minor changes that I only noticed after the CI failures.

charles-lunarg (approved):
You definitely addressed the issues I have with the original draft so thank you for doing that. Seems like there needs to be a special incantation to get the module to compile in the first place, but I’m unsure of what that would be as I haven’t tried getting modules to work with cmake myself.

sharadhr:
> Seems like there needs to be a special incantation to get the module to compile in the first place

It's probably `cmake --build path/to/build/dir` to _actually_ compile the module; however, built module interfaces (BMIs) are not portable and should be compiled by the consumer, as suggested by the CMake repository. 

It's strange that that particular CI step is [failing here](https://github.com/KhronosGroup/Vulkan-Headers/actions/runs/9523347655/job/26254620665), but passes in [my repository](https://github.com/sharadhr/Vulkan-Headers/actions/runs/9523356687/job/26254629887). With CMake 3.28, this should be entirely automatic with zero additional interaction by the user.

oddhack:
> Changes look appropriate for what is being done - I would suggest squashing the commits so that the intermediate steps aren't in the git history. Same with the github PR description being out of date with the current set of changes.

We can configure automatic squash-on-merge for this repo if you like.

charles-lunarg:
> It's strange that that particular CI step is [failing here](https://github.com/KhronosGroup/Vulkan-Headers/actions/runs/9523347655/job/26254620665), but passes in [my repository](https://github.com/sharadhr/Vulkan-Headers/actions/runs/9523356687/job/26254629887). With CMake 3.28, this should be entirely automatic with zero additional interaction by the user.

Looking at your fork's CI run, it appears that it didn't run this branch, but only main. Hence it passed with flying colors since it wasn't trying to enable the c++ module.

GCC-11 and Clang-15 (apple clang) both are listed as not supporting the C++ Module Scanning that CMake requires, so I believe the best course of action is adding new CI jobs for newer compilers as well as disable the module support if the compiler isn't new enough in the other runs.

charles-lunarg:
> > Changes look appropriate for what is being done - I would suggest squashing the commits so that the intermediate steps aren't in the git history. Same with the github PR description being out of date with the current set of changes.
> 
> We can configure automatic squash-on-merge for this repo if you like.

I believe the current developer policy is rebasing which I prefer. Asking users to squish their commits if they so desire is suitable in my opinion.

sharadhr:
> Looking at your fork's CI run, it appears that it didn't run this branch, but only main. Hence it passed with flying colors since it wasn't trying to enable the c++ module.

Ah, I just noticed that myself. I've modified the guard to test for the compiler and compiler version. I'm not quite sure how to modify the CI without dragging in huge from-source set-ups for latest Clang and GCC.

charles-lunarg:
I just did some additional testing myself, and the CI fails because installing the headers on Windows-latest fails. 
Gist is that its trying to install `Vulkan-Module.lib` which doesn't exist cause its not a regular library. 
Doing some research about how to install modules, there may not be a consensus on how to do that just yet. Vulkan-Hpp doesn't appear to even try, it just installs `vulkan.cppm` as a header file and leaves it as that.

So I went ahead and branched on your change to test some stuff out. Adding an explicit install step revealed the underlying issue, and that is fixed by just not trying to install Vulkan-Module. Not sure if that is a solution, or just a workaround.

Regardless, while 2 of the 3 tests pass, the add_subdirectory test is still failing with an internal compiler error, which makes it quite difficult to know "what is wrong". https://github.com/charles-lunarg/Vulkan-Headers/actions/runs/9569931247/job/26383560260

charles-lunarg:
Good news is that the internal compiler error isn't due to running in the integration test. I realized the CI job needs to do an explicit build step now that the module may require explicit building.

https://github.com/charles-lunarg/Vulkan-Headers/actions/runs/9570194204/job/26384402318

Still an internal compiler error, but figured that should be said since it'll help figure it out.

sharadhr:
Ah, that ICE [is known and reported](https://developercommunity.visualstudio.com/t/On-building-a-C20-module:-fatal--error/10469799), and should be fixed in VS 2022 17.11.

charles-lunarg:
Okay that is GREAT news!

A known ICE with a "fix is coming" means I will go ahead and fixup the PR to do the requisite CMake changes, like adding a build option (which defaults to on, but can be set to OFF in case downstream users see issues like we did here). For the time being, windows has the module option set to OFF so that CI passes. Future work is to turn it on when MSVC 17.11 makes it into the runner. 

Part of the changes is to remove the installing of the vulkan-module for the time being. It appears, after doing some research, that how to install C++ modules is still being worked on. It's prudent to air on the side of "don't" while the CMake devs work on documentation on exactly how to do it. At least in this PR the module is built & usable for users using add_subdirectory. 

I also went ahead and added build & install steps to the CI - making it far easier to see where the build/install goes wrong.

charles-lunarg (approved):
Approving again now that there is a large set of changes since the previous approval.

sharadhr:
Cheers; I appreciate the additional work done to get the CI happy!

PR #487, opened 2024-06-22
Makefile to import generated scripts and headers when doing a spec update

oddhack:
Motivated by recent problems building the SC 1.0.15 spec update - will import to VulkanSC-Headers and tweak once it's proven out for Vulkan-Headers.

The Makefile may need to be modified when directory structure is changed or new scripts or headers are added.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,108 @@
+# Copyright 2024 The Khronos Group Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+# Makefile to update external files generated in other repositories when
+# a public specification update is done.
+
+# Needed to get the right version of test, apparently
+SHELL  = /bin/bash
+
+REVISION = 999
+
+# Location of other repository clones
+GIT	 = ..
+SPEC	 = $(GIT)/Vulkan-Docs
+HPP	 = $(GIT)/Vulkan-Hpp
+REGISTRY = $(GIT)/registry/vulkan
+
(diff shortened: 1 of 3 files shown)


charles-lunarg (commented):
Is there a particular reason this is a Makefile and not just another python script?

I worry that the use of makefile will give the wrong impression to users. If I saw a project with a makefile, I would naively assume that its meant for 'building the project' (which isn't even necessary). Whereas the actual makefile added appears to be a utility script for header/spec updates.

oddhack:
make is the right tool for the task and python is not.

It doesn't have to be *named* 'Makefile' if that will assuage the concern.

oddhack:
N.b. if 'building the project' isn't necessary, the same consideration would apply to the cmake stuff IMO. The BUILD.md file literally starts with 'Instructions for building this repository'.

charles-lunarg:
Ah, the filename was probably my biggest concern. So changing that would assuage all of my concern.

charles-lunarg:
My "building isn't necessary" is referencing how this is a header only project. Although now it is *mostly* header only, with vulkan.cppm blurring the lines. A user could clone/download the Vulkan-Headers repo, point their compiler to the include folder and pretty much be done. So while CMake is indeed used to 'build' the project, it isn't building binaries, rather it is setting up a CMake target for downstream users as well as managing the `install` step for package managers.

charles-lunarg (approved):

PR #492, opened 2024-07-12
Disable VulkanHppModule support in clang-cl 

charles-lunarg:
The current released versions of clang-cl does not work with CMakes
module implementation. Rather than break downstream users, just disable
the module support for the time being.


.github/workflows/ci.yml
```diff
@@ -52,6 +52,27 @@ jobs:
       - run: ctest --output-on-failure
         working-directory: build
 
+  windows_clang:
+    runs-on: windows-2022
+    strategy:
+      matrix:
+          compiler: [ clang, clang-cl ]
+          config: [ Release ]
```
mikes-lunarg, line 60: You aren't using the config parameter in the steps

charles-lunarg: Oh right, I'll fix that.

mikes-lunarg (approved):

mikes-lunarg:
FYI: This issue affects _ALL_ users of clang-cl, whether or not they intend to use C++ modules. Since the 290 header will see wider distribution as an SDK release, @charles-lunarg and I decided to move the `v1.3.290` tag to include this fix and avoid more downstream breakage.

PR #501, opened 2024-08-30
Fix MSVC identification for modules support

### LINKED ISSUE #500: Windows cmake CI failing
Something started failing in some of the windows-clang jobs in today's spec update. This looks like maybe the underlying clang version changed in an incompatible way. I can't assess this and can't hold up the spec release to wait for it to be fixed, so went ahead and merged the 1.3.295 update. The Linux clang jobs work fine.

See https://github.com/KhronosGroup/Vulkan-Headers/actions/runs/10630554267/job/29469652043#step:4:82

mikes-lunarg:
Can't use `MSVC` var because it is also true for clang-cl. Fixes #500.


charles-lunarg (approved):
Thank you! I was confused why this started failing, but clang-cl breaking all of the common assumptions is par for the course.

PR #511, opened 2024-11-01
Undefine major and minor through <sys/sysmacros.h>

gulfemsavrun:
<sys/sysmacros.h> is included through some other headers. This results in major(x) being resolved to gnu_dev_major(x) which is an expression in a constructor initializer list. This patch includes <sys/sysmacros.h> at the very beginning, and undefines major and minor to avoid them to be resolved to gnu_dev_major(x) or gnu_dev_minor(x).

### DIFF
--- a/include/vulkan/vulkan.hpp
+++ b/include/vulkan/vulkan.hpp
@@ -8,6 +8,13 @@
 #ifndef VULKAN_HPP
 #define VULKAN_HPP
 
+// <sys/sysmacros.h> is included through some other headers, and
+// this results in major(x) being resolved to gnu_dev_major(x)
+// which is an expression in a constructor initializer list.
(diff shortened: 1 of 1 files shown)


charles-lunarg:
Unfortunately, I cannot accept the PR because this repo only contains copies of the header files, it is not the source for them as documented in [CONTRIBUTING.md.](https://github.com/KhronosGroup/Vulkan-Headers/blob/main/CONTRIBUTING.md)

The Vulkan-Hpp header (and now module) files are in https://github.com/KhronosGroup/Vulkan-Hpp. Any changes to those files must go to that repository.

oddhack:
@charles-lunarg you can also move issues to other KhronosGroup repos (or ask me to, if you don't have the right permissions).

PR #512, opened 2024-11-02
Update for Vulkan-Docs 1.3.301

oddhack:
Note that this update populates the "page" key in validusage.json with the relative path to the HTML page containing that VUID on docs.vulkan.org (or any other Antora build) in the spec/latest module.


MarijnS95:
@oddhack just checking, it looks like the `v1.3.301` tag is missing: https://github.com/KhronosGroup/Vulkan-Headers/tags

Issue #517, opened 2024-12-04
validusage.json not updated with 1.4 version


spencer-lunarg:
@oddhack it seems when @arno-lunarg / @mikes-lunarg were trying to grab the new 1.4 Vulkan Headers that `validusage.json` was never updated (https://github.com/KhronosGroup/Vulkan-Headers/commits/main/registry/validusage.json)

Is there a way we can updated it so that we can pull it in for Validation

spencer-lunarg:
Made a PR, mainly because this is blocking Validation from being merged (because the implicit VUs names are wrong and fail our tests) so made a branch we can work on (please feel free to close my PR with a real fix, mainly because Github doesn't even let you view the diff online!)

oddhack:
It did get generated and checked into the registry, but the Makefile.release that adds it to the Vulkan-Headers updates needed a minor tweak for the new registry structure. So the right validusage.json is in HEAD now, but the 1.4.303 tag does not point there, yet. I can go ahead and update the tag to point to current HEAD if that won't further complicate matters.

spencer-lunarg:
>  but the 1.4.303 tag does not point there, yet. I can go ahead and update the tag to point to current HEAD if that won't further complicate matters.

we can point to a direct commit

oddhack:
I mean, the tag should point to the right file anyway. If updating it will not affect your process  not I will go ahead and do it. In any event current main branch contains what should be the right validusage now, or at least, the one built out of the 1.4.303 tag in Vulkan-Docs.

oddhack:
Tag has been updated. Please close once you've verified the validusage.json is good.

arno-lunarg:
Working now, thank you!

PR #518, opened 2024-12-04
Update validusage.json with 1.4

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-Headers/issues/517


oddhack:
Updated validusage.json has already been committed to main branch.

PR #522, opened 2025-01-21
ubuntu arm runners

johnnynunez:


charles-lunarg:
Closing due to the lack of need - Vulkan-Headers repo doesn't "build" anything, the CI is just to check that the CMake code works correctly. Adding ARM costs minutes and doesn't benefit anything.

PR #528, opened 2025-02-17
CMake: Cleanup of `Vulkan-HppModule` target

M2-TE:
The `Vulkan-Hpp` headers have numerous compile definitions for configuration, which can of course be done manually, but it is much nicer to have actual support for them in CMake via `option()`.
This is particularly true for the C++20 module, which requires compile definitions at the configure stage due to the lack of preprocessor stuff with `#include`.
Projects relying on Vulkan (e.g. the VulkanMemoryAllocator) can also consume these propagating compile definitions to not have to define them multiple times.

To this end, I made some changes to the `CMakeLists.txt`, similar to what is now in the `Vulkan-Hpp` repository, to provide three separate targets with consolidated naming:

- `Vulkan-Headers ALIAS Vulkan::Headers`: Same as before, so it will not break anything.
- `Vulkan-Hpp ALIAS Vulkan::Hpp`: Separated from `Vulkan-Headers`, as to provide isolated compile definitions for this C++ target, even though it shares the same interface include directory as the C target.
- `Vulkan-HppModule ALIAS Vulkan::HppModule`: Consumes `Vulkan-Hpp` and its compile definitions. Adjusted naming to match the names of the repository and other targets. Added errors for mismatching compilers (previously warning) and CMake versions. Added ability to build the C++20 module with `import std;` (as this is still a very experimental feature for most compilers, I have not updated the compiler requirements for it).

Currently, the `VK_NO_PROTOTYPERS` and `VULKAN_HPP_DISPATCH_LOADER_DYNAMIC` options are separated as per their respective compile definitions, but it might make sense to consolidate both into a single option to enable the respective C/C++ flag in the two targets.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,42 +38,67 @@ vlk_get_header_version()
 
 project(VULKAN_HEADERS LANGUAGES C CXX VERSION ${VK_VERSION_STRING})
 
+# options for Vulkan-Headers and the Vulkan-Hpp C++20 module
+option(VULKAN_HEADERS_ENABLE_TESTS "Test Vulkan-Headers" ${PROJECT_IS_TOP_LEVEL})
+option(VULKAN_HEADERS_ENABLE_INSTALL "Install Vulkan-Headers" ${PROJECT_IS_TOP_LEVEL})
+option(VULKAN_HEADERS_ENABLE_MODULE "Enables building of the Vulkan C++20 module; requires minimum CMake version 3.28" OFF)
+option(VULKAN_HEADERS_ENABLE_MODULE_STD "Enables building of the Vulkan C++20 module with import std; requires minimum CMake version 3.30" OFF)
+
+# set up Vulkan-Headers
 add_library(Vulkan-Headers INTERFACE)
 add_library(Vulkan::Headers ALIAS Vulkan-Headers)
 target_include_directories(Vulkan-Headers INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>)
 
-if ((CMAKE_CXX_COMPILER_ID STREQUAL "MSVC" AND (MSVC_VERSION GREATER_EQUAL "1941")) OR
-    # clang-cl doesn't currently support modules
-    (CMAKE_CXX_COMPILER_ID STREQUAL "Clang"
-        AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "16.0"
-        AND (NOT CMAKE_C_COMPILER_FRONTEND_VARIANT MATCHES "MSVC")
-        AND (NOT CMAKE_CXX_COMPILER_CLANG_SCAN_DEPS STREQUAL CMAKE_CXX_COMPILER_CLANG_SCAN_DEPS-NOTFOUND)) OR
-    (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "14.0"))
-    set(COMPILER_SUPPORTS_CXX_MODULES TRUE)
-endif()
-
-option(VULKAN_HEADERS_ENABLE_MODULE "Enables building of the Vulkan C++ module" OFF)
-
-if (VULKAN_HEADERS_ENABLE_MODULE AND (NOT COMPILER_SUPPORTS_CXX_MODULES OR CMAKE_VERSION VERSION_LESS "3.28"))
-    message(WARNING "Vulkan C++ module support is requested but was disabled due to lacking support on this platform")
-endif()
-
-if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.28" AND VULKAN_HEADERS_ENABLE_MODULE AND COMPILER_SUPPORTS_CXX_MODULES)
-    add_library(Vulkan-Module)
-    add_library(Vulkan::VulkanHppModule ALIAS Vulkan-Module)
-    target_sources(Vulkan-Module
-        PUBLIC
-            FILE_SET module
-                TYPE CXX_MODULES
-                BASE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}/include"
-                FILES "${CMAKE_CURRENT_SOURCE_DIR}/include/vulkan/vulkan.cppm"
+if (VULKAN_HEADERS_ENABLE_MODULE)
+    # check for compiler support
+    if ((CMAKE_CXX_COMPILER_ID STREQUAL "MSVC" AND (MSVC_VERSION GREATER_EQUAL "1941")) OR
+        # clang-cl doesn't currently support modules
+        (CMAKE_CXX_COMPILER_ID STREQUAL "Clang"
+            AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "16.0"
+            AND (NOT CMAKE_C_COMPILER_FRONTEND_VARIANT MATCHES "MSVC")
+            AND (NOT CMAKE_CXX_COMPILER_CLANG_SCAN_DEPS STREQUAL CMAKE_CXX_COMPILER_CLANG_SCAN_DEPS-NOTFOUND)) OR
+        (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "14.0")
     )
-    target_compile_features(Vulkan-Module PUBLIC cxx_std_20)
-    target_link_libraries(Vulkan-Module PUBLIC Vulkan-Headers)
-endif ()
+        # check for CMake support
+        if(VULKAN_HEADERS_ENABLE_MODULE_STD AND CMAKE_VERSION VERSION_LESS "3.30")
+            message(FATAL_ERROR "Vulkan-Hpp: C++20 module with import std requires CMake 3.30 or later")
+        elseif (CMAKE_VERSION VERSION_LESS "3.28")
+            message(FATAL_ERROR "Vulkan-Hpp: C++20 module requires CMake 3.28 or later")
+        endif()
+        
+        # set up Vulkan-HppModule
+        add_library(Vulkan-HppModule)
+        add_library(Vulkan::HppModule ALIAS Vulkan-HppModule)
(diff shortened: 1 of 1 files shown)


jpr42 (commented):
Projects who consume Vulkan-Headers via add_subdirectory and FetchContent will now have  their CMakeCache.txt flooded with cache variables that IMO don't add much.

Seems much simpler for projects to set compile definitions on their own.

I like the cleanup to the VULKAN_HPP_ENABLE_MODULE module code though.

jpr42:
> The only ones that might be worth keeping would be the VK_NO_PROTOTYPES and VULKAN_HPP_DISPATCH_LOADER_DYNAMIC compile definitions for the C and C++ headers respectively, which could both be uniformly controlled via a VULKAN_HEADERS_DYNAMIC CMake variable or something similar? What do you think?

Since they ultimately  boil down to compile definitions I still don't think it's worth it. 

Furthermore the CMake code you proposed here:
```
if (VULKAN_HPP_DISPATCH_LOADER_DYNAMIC)
	target_compile_definitions(Vulkan-Hpp INTERFACE "VULKAN_HPP_DISPATCH_LOADER_DYNAMIC")
    if (NOT VK_NO_PROTOTYPES)
        message(WARNING "VK_NO_PROTOTYPES disabled while using VULKAN_HPP_DISPATCH_LOADER_DYNAMIC, consider enabling both")
    endif()
endif()
```

Could also be written as C++ code in Vulkan-Hpp using the [preprocessor](https://en.cppreference.com/w/cpp/preprocessor).

```
#if defined(VULKAN_HPP_DISPATCH_LOADER_DYNAMIC) && !defined(VK_NO_PROTOTYPES)
#warning  "VK_NO_PROTOTYPES disabled while using VULKAN_HPP_DISPATCH_LOADER_DYNAMIC, consider enabling both"
#endif
```

Maybe a VULKAN_HEADERS_DYNAMIC is a good idea for a compile definition, but that's up to the Vulkan-Hpp authors. This is more of a library question for them.

> With the removal of the HPP options, Vulkan::Hpp will essentially just be an interface for Vulkan::Headers. I think it is still very much worth keeping around for clarity sake and as a dependency for the module target. Do you still think I should add tests for it?

I don't see too much value in the Vulkan:Hpp target all above discussion considered.

Furthermore, the testing in this repo does integration testing. Making sure add_subdirectory / find_package users can consume the library without issues. Even though it's overkill testing it's important to have. I added it precisely because it's important that never breaks. Changing a target name is a breaking change. The integration testing in this repo should catch that.

M2-TE:
I've removed the additional `Vulkan::Hpp` target alongside the additional options, while leaving the cleanup of the C++20 module.

As far as I can tell, the integration tests lack testing for said module:

https://github.com/KhronosGroup/Vulkan-Headers/blob/952f776f6573aafbb62ea717d871cd1d6816c387/tests/CMakeLists.txt#L10-L16

and, based on what you said, it might be worth adding (albeit in a separate PR).

CMakeLists.txt
```diff
@@ -38,42 +38,58 @@ vlk_get_header_version()
...
+        # set up Vulkan-HppModule
+        add_library(Vulkan-HppModule)
+        add_library(Vulkan::HppModule ALIAS Vulkan-HppModule)
+        target_sources(Vulkan-HppModule
+            PUBLIC
+            FILE_SET module
+            TYPE CXX_MODULES
+            FILES "${CMAKE_CURRENT_SOURCE_DIR}/include/vulkan/vulkan.cppm")
+        set_target_properties(Vulkan-HppModule PROPERTIES
+            CXX_STANDARD_REQUIRED ON # CXX version is a hard requirement for modules
```
jpr42, line 78: Is this not enforced by compile features? The official example also doesn't have it.

https://www.kitware.com/import-std-in-cmake-3-30/

CMakeLists.txt
```diff
@@ -38,42 +38,58 @@ vlk_get_header_version()
...
+            CXX_EXTENSIONS OFF) # Clang's module support can be broken with extensions enabled
```
jpr42, line 79: Do you have a link to an issue for this?

CMakeLists.txt
```diff
@@ -38,42 +38,58 @@ vlk_get_header_version()
 
 project(VULKAN_HEADERS LANGUAGES C CXX VERSION ${VK_VERSION_STRING})
 
+# options for Vulkan-Headers and the Vulkan-Hpp C++20 module
+option(VULKAN_HEADERS_ENABLE_TESTS "Test Vulkan-Headers" ${PROJECT_IS_TOP_LEVEL})
+option(VULKAN_HEADERS_ENABLE_INSTALL "Install Vulkan-Headers" ${PROJECT_IS_TOP_LEVEL})
+option(VULKAN_HEADERS_ENABLE_MODULE "Enables building of the Vulkan C++20 module; requires minimum CMake version 3.28" OFF)
+option(VULKAN_HEADERS_ENABLE_MODULE_STD "Enables building of the Vulkan C++20 module with import std; requires minimum CMake version 3.30" OFF)
```
jpr42, line 45: OFF is the default.

```suggestion
option(VULKAN_HEADERS_ENABLE_MODULE "Enables building of the Vulkan C++20 module; requires minimum CMake version 3.28")
option(VULKAN_HEADERS_ENABLE_MODULE_STD "Enables building of the Vulkan C++20 module with import std; requires minimum CMake version 3.30")
```

jpr42:
Please also rebase this into 1 commit.

CMakeLists.txt
M2-TE: Is it not better to explicitly set it to OFF for the sake of readability? I've seen this approach in pretty much all CMake files so far. If you still prefer to omit it, I'll of course oblige

jpr42: FFTI this bit of feedback. Minor nitpick from me.

CMakeLists.txt
M2-TE: I cannot find the discrete issue on this, but it was referenced in the official example for C++ modules in 2.28:
[https://www.kitware.com/import-cmake-the-experiment-is-over/](https://www.kitware.com/import-cmake-the-experiment-is-over/)

Specifically, it seems to be an issue with Clang 16 only, so it might be worth checking for the clang version here and only conditionally disabling `CXX_EXTENSIONS`?

CMakeLists.txt
M2-TE: You are absolutely right, I will remove it

CMakeLists.txt
jpr42: Feel free to ignore.

M2-TE:
> Please also rebase this into 1 commit.

Are the commits not squashed automatically when merged into main? If not, how would I best go about doing this?
As for the testing, I will add it in a separate PR!

jpr42:
> Are the commits not squashed automatically when merged into main?

Not automatically. But yes it can be done. FFTI

jpr42 (approved):
FYI @charles-lunarg

charles-lunarg (approved):
Changes look great, thanks @jpr42 for the extensive review!

CMakeLists.txt
```diff
@@ -38,42 +38,60 @@ vlk_get_header_version()
...
+        if(VULKAN_HEADERS_ENABLE_MODULE_STD AND CMAKE_VERSION VERSION_LESS "3.30")
+            message(FATAL_ERROR "Vulkan-Hpp: C++20 module with import std requires CMake 3.30 or later")
+        elseif (CMAKE_VERSION VERSION_LESS "3.28")
+            message(FATAL_ERROR "Vulkan-Hpp: C++20 module requires CMake 3.28 or later")
+        endif()
+        
+        # set up Vulkan-HppModule
+        add_library(Vulkan-HppModule)
+        add_library(Vulkan::HppModule ALIAS Vulkan-HppModule)
+        target_sources(Vulkan-HppModule
```
charles-lunarg, line 72: Hang on, why was Vulkan-Module renamed? There is a comment in the PR about adding an alias for Vulkan-Module but I do not see one in the source code anywhere. Did this get dropped by mistake?

charles-lunarg:
I do not mind squashing the commit & fixing up the message before merging. I started working on that before noticing the above issue (rename Vulkan-Module to Vulkan-HppModule without an alias), will hold off squashing until that is resolved.

CMakeLists.txt
M2-TE: I failed to explain this properly; I renamed Vulkan-Module to Vulkan-HppModule to better hint at these being the hpp headers and also fit the naming scheme of the previously present Vulkan-Hpp target that was dropped later on.

Should I rename this back to `Vulkan-Module` and `Vulkan::Module` or add those two as _additional_ alias targets?

charles-lunarg: I do think that the Vulkan-HppModule is a better name than just Vulkan-Module, but I also don't want to break people who are using Vulkan-Module currently.

How about adding in aliases but adding a deprecation notice as well? So that in 6-12 months we can remove the old names without suddenly breaking downstream users?

```cmake
add_library(Vulkan-Module ALIAS Vulkan-VulkanHppModule)
add_library(Vulkan::Module ALIAS Vulkan-VulkanHppModule)
message(DEPRECATION "The Vulkan-Module and Vulkan::Module targets have been deprecated by  Vulkan-VulkanHppModule and will be removed at a future date.")
``` 

By putting the above cmake code inside of the `if (VULKAN_HEADERS_ENABLE_MODULE)` block, the deprecation notice should only appear to users of Vulkan-Module. I looked but was unable to find a way to mark a CMake target as deprecated, so this was the alternative I came up with.

M2-TE: I found this in the CMake docs:
[https://cmake.org/cmake/help/latest/prop_tgt/DEPRECATION.html](https://cmake.org/cmake/help/latest/prop_tgt/DEPRECATION.html)

And just tested it myself. Setting the `DEPRECATION` property of a target to a string seems to warn the user about the deprecation and prints out said string, whenever they try to link against it. Tested it in CMake 3.31 but as per docs, should work in 3.17 and up!

M2-TE: Alright, I have added `Vulkan-Module` as an `INTERFACE` library, as downstream users may be using it as such. Setting it as an ALIAS might break things for them (setting custom properties and compile defs would not work, for example). The `Vulkan::Module` is an ALIAS target, just like before and both targets should now behave exactly as they did before with the addition of the deprecation notice.

PR #536, opened 2025-03-18
ci: Integration tests for C++20 modules

M2-TE:
Additions:

- Added a separate integration test
- Added checks for CMake target availability

Current issues:

- I do not know which environments you run these tests in apart from CI, but modules have pretty modern requirements, e.g. CMake 3.28+, gcc 14+ or clang 16+ (both preferrably higher) on linux. The tests will likely fail when the .cppm cannot be compiled, especially since some CI setups do not meet these requirements. Should I remedy this via compiler and cmake version checks within the test itself?
- find_package testing currently does not test for modules, as `find_package(VulkanHeaders CONFIG)` will not make it available afaik, at least on my local machine. Where does this CMake file live?
- There is currently a bug in `vulkan.cppm` (fixed upstream in KhronosGroup/Vulkan-Hpp#2099) that will prevent the module tests from succeeding.


tests/CMakeLists.txt
```diff
@@ -6,15 +6,24 @@
...
-# Test add_subdirectory suppport
-add_test(NAME integration.add_subdirectory
+# Test add_subdirectory suppport for headers
+add_test(NAME integration.add_subdirectory_headers
     COMMAND ${CMAKE_CTEST_COMMAND}
         --build-and-test ${CMAKE_CURRENT_LIST_DIR}/integration
-                         ${CMAKE_CURRENT_BINARY_DIR}/add_subdirectory
+                         ${CMAKE_CURRENT_BINARY_DIR}/add_subdirectory_headers
         --build-generator ${CMAKE_GENERATOR}
         --build-options -DFIND_PACKAGE_TESTING=OFF -DVULKAN_HEADERS_ENABLE_MODULE=OFF
```
jpr42, line 15: I don't see a reason for 2 add_subdirectory tests.

```suggestion
        --build-options -DFIND_PACKAGE_TESTING=OFF -DVULKAN_HEADERS_ENABLE_MODULE=${VULKAN_HEADERS_ENABLE_MODULE}
```

jpr42 (changes requested):

tests/integration/CMakeLists.txt
jpr42, line 1: Testing the target name isn't really enough. An executable linking and using Vulkan::HppModule is still needed.

jpr42:
> I do not know which environments you run these tests in apart from CI, but modules have pretty modern requirements, e.g. CMake 3.28+, gcc 14+ or clang 16+ (both preferrably higher) on linux. The tests will likely fail when the .cppm cannot be compiled, especially since some CI setups do not meet these requirements.

These tests really only get run in github actions, see .github/workflows/ci.yml

For these kind of things you got to see what the runners have available:
https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md
https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md

> Should I remedy this via compiler and cmake version checks within the test itself?

Adding a new CI test dedicated toward modules seems the best way to go IMO.

Example (Note I didn't test this, but something like this should work):

```
  cmake-linux-module:
    runs-on: ubuntu-24.04
    steps:
      - uses: actions/checkout@v4
      - uses: lukka/get-cmake@latest
      - run: |
          cmake -S . -B build \
         -D VULKAN_HEADERS_ENABLE_TESTS=ON -D VULKAN_HEADERS_ENABLE_INSTALL=ON \
         -G Ninja \
         -D VULKAN_HEADERS_ENABLE_MODULE=ON \
         -D VULKAN_HEADERS_ENABLE_MODULE_STD=ON \
         -D CMAKE_CXX_COMPILER=g++-14 \
         -D CMAKE_C_COMPILER=gcc-14
      - run: cmake --build ./build
      - run: cmake --install build/ --prefix build/install
      - run: ctest --output-on-failure
        working-directory: build
```

> find_package testing currently does not test for modules, as find_package(VulkanHeaders CONFIG) will not make it available afaik, at least on my local machine. Where does this CMake file live?

Yes currently the module target is not being installed. No one has written it yet. That's another PR separate from this one.

> There is currently a bug in vulkan.cppm (fixed upstream in https://github.com/KhronosGroup/Vulkan-Hpp/pull/2099) that will prevent the module tests from succeeding.

Seems like Vulkan-Hpp needs more testing. Those kinds of bugs shouldn't make it here.

jpr42:
IMO we shouldn't add tests here yet until we have more testing in Vulkan-Hpp.

jpr42:
Also @M2-TE do you know if C++20 modules work on Apple at all?

tests/CMakeLists.txt
M2-TE: I separated them since I did not yet know how to best filter out module support testing for certain CI envs, passing the var like you did here seems like the best solution!

tests/integration/CMakeLists.txt
M2-TE: Agreed, the module still gets compiled even when not linked, but that sort of testing will be done in the upstream repo anyways. Should I add a .cpp file to testing/ or is there a way to link to a dummy target?

M2-TE:
> Yes currently the module target is not being installed. No one has written it yet. That's another PR separate from this one.

Got it, I will omit find_package testing for now.

> Seems like Vulkan-Hpp needs more testing. Those kinds of bugs shouldn't make it here. IMO we shouldn't add tests here yet until we have more testing in Vulkan-Hpp.

I agree, we are in the process of adding proper testing for these modules upstream, which should be done by the time the fixed .cppm gets merged here. It's the main reason I marked this as draft currently.

> Also M2-TE do you know if C++20 modules work on Apple at all?

@jpr42 I am not quite sure yet, since I have no hardware to test this on. I will use CI to test this in the Vulkan-Hpp repo first, but it should work, since Apple uses Clang too afaik. Supported versions for modules are clang++-16 to 19 atm.

tests/integration/CMakeLists.txt
jpr42: > Should I add a .cpp file to testing/ or is there a way to link to a dummy target?

Add a cpp file in the `tests` directory.

A very minimal example importing vulkan.cppm

M2-TE:
I've added minimal cpp libraries to test both the `Vulkan::Hpp` and `Vulkan::HppModules` targets.
The two additional CI workflows will handle module testing, specifically using clang++18 under Ubuntu and latest MSVC under Windows. Macos will not be possible with the current runners.

Notably, `VULKAN_HEADERS_ENABLE_MODULE_STD` should not be tested yet, as it is still very much an experimental feature in CMake.

tests/vk_hpp.cpp
```diff
@@ -0,0 +1,6 @@
+#include <vulkan/vulkan.hpp>
```
jpr42, line 1:
```suggestion
/*
 * Copyright 2025 The Khronos Group Inc.
 * Copyright 2025 Valve Corporation
 * Copyright 2025 LunarG, Inc.
 *
 * SPDX-License-Identifier: Apache-2.0
 */
#include <vulkan/vulkan.hpp>
```

tests/vk_hpp_module.cpp
```diff
@@ -0,0 +1,6 @@
+import vulkan_hpp;
```
jpr42, line 1:
```suggestion
/*
 * Copyright 2025 The Khronos Group Inc.
 * Copyright 2025 Valve Corporation
 * Copyright 2025 LunarG, Inc.
 *
 * SPDX-License-Identifier: Apache-2.0
 */
import vulkan_hpp;
```

tests/vk_hpp.cpp
```diff
@@ -0,0 +1,6 @@
...
+
+int square(int i)
+{
+    return i * i;
+}
```
jpr42, line 6:
```suggestion
int header_version()
{
    return VK_HEADER_VERSION;
}
```

tests/vk_hpp_module.cpp
```diff
@@ -0,0 +1,6 @@
...
+
+int square(int i)
+{
+    return i * i;
+}
```
jpr42, line 6:
```suggestion
int header_version()
{
    return VK_HEADER_VERSION;
}
```

tests/integration/CMakeLists.txt
```diff
@@ -71,3 +77,16 @@ target_link_libraries(vk_icd PRIVATE Vulkan::Headers)
...
+
+# vulkan.hpp
+add_library(vk_hpp MODULE ../vk_hpp.cpp)
+target_compile_features(vk_hpp PUBLIC cxx_std_11)
+target_link_libraries(vk_hpp PRIVATE Vulkan::Headers)
+
+# vulkan.cppm
+if (VULKAN_HEADERS_ENABLE_MODULE)
+    add_library(vk_hpp_module MODULE ../vk_hpp_module.cpp)
+    target_compile_features(vk_hpp_module PUBLIC cxx_std_20)
```
jpr42, line 89: This line should not be needed. Linking against Vulkan::HppModule should automatically handle this.

`Vulkan::HppModule` sets `cxx_std_20` when they link against the library.

```suggestion
```

M2-TE:
Module tests now work properly with the 1.4.311 `vulkan.cppm` source.

For future reference: Is there a way for me to test with CI without pushing? The commits may get merged anyways, but I don't want to create clutter that makes it more annoying for others to go through the specific changes I made.

jpr42:
> For future reference: Is there a way for me to test with CI without pushing?

Haven't tried it myself but I know some devs use this to run github actions locally: https://github.com/nektos/act

charles-lunarg (approved):
Nothing here gives me pause. I haven't read through the full discussion that was had (I was on PTO at the time). 

I do notice that this is still marked as a draft, are there still changes that are wanting to be made?

charles-lunarg:
> The macos runners seem to only go up to Clang 15 (in both macos-14 and macos-15), so these can't be used to test modules yet: https://github.com/actions/runner-images/tree/main/images/macos

The link provided states that clang-18 is available from homebrew, `$(brew --prefix llvm@18)/bin/clang`

M2-TE:
@charles-lunarg
> are there still changes that are wanting to be made?

There are no changes still to be made, but I am waiting for upstream module testing in Vulkan-Hpp to be merged first.

> The link provided states that clang-18 is available from homebrew, $(brew --prefix llvm@18)/bin/clang

I missed that in the readme, I will try to get this to work in Vulkan-Hpp and add it to a runner here if it does!

M2-TE:
@charles-lunarg 
Module testing is now merged in Vulkan-Hpp, meaning that this can be merged as well.

I did not manage to get it to compile under macos-15 with clang++-18 and the errors with modules can sometimes be a bit difficult to hunt down.. I don't personally have a mac so debugging purely with CI sounds like a bit of a nightmare and until there is stable testing for the vulkan.cppm module under macos, it should not get any integration tests here I think.

jpr42:
Thanks for the hard work @M2-TE it's much appreciated.

M2-TE:
Thank you for the thorough reviews as well @jpr42 !

PR #538, opened 2025-04-02
CMake: C++20-Module installation and integration testing

M2-TE:
- Installing Vulkan-HppModule target with Vulkan::HppModule ALIAS
- FindPackage testing for module target
- Cleaned up minimal .cpp
- Properly utilized the existing Ninja variable in CMAKE_GENERATOR within CI


CMakeLists.txt
```diff
@@ -115,7 +115,12 @@ if (VULKAN_HEADERS_ENABLE_INSTALL)
     install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/registry" DESTINATION "${CMAKE_INSTALL_DATADIR}/vulkan" USE_SOURCE_PERMISSIONS)
 
     set_target_properties(Vulkan-Headers PROPERTIES EXPORT_NAME "Headers")
-    install(TARGETS Vulkan-Headers EXPORT VulkanHeadersConfig INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+    if (VULKAN_HEADERS_ENABLE_MODULE)
+        set_target_properties(Vulkan-HppModule PROPERTIES EXPORT_NAME "HppModule")
+        install(TARGETS Vulkan-Headers Vulkan-HppModule EXPORT VulkanHeadersConfig INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+    else()
+        install(TARGETS Vulkan-Headers EXPORT VulkanHeadersConfig INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+    endif()
```
charles-lunarg, line 123: Logic might be simpler to read if there was a list of install targets

```suggestion
    list(APPEND INSTALL_TARGETS Vulkan-Headers)
    if (VULKAN_HEADERS_ENABLE_MODULE)
            set_target_properties(Vulkan-HppModule PROPERTIES EXPORT_NAME "HppModule")
            list(APPEND INSTALL_TARGETS Vulkan-HppModule)
    endif()
    install(TARGETS ${INSTALL_TARGETS} EXPORT VulkanHeadersConfig INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
```

M2-TE: That looks a lot better, you're right!

M2-TE:
> CMake Error at CMakeLists.txt:123 (install):
  install TARGETS target Vulkan-HppModule is exported but not all of its
  interface file sets are installed

I looks like cmake wants us to explicitly install the .cppm file as well, even though we already install it implicitly here

https://github.com/KhronosGroup/Vulkan-Headers/blob/2ac81691baf291e7f4aad07596d7073974dbc4dd/CMakeLists.txt#L113

CMake `install` lists a property specifically for cxx modules as [CXX_MODULES_BMI](https://cmake.org/cmake/help/latest/command/install.html#signatures), so I will try to use that

M2-TE:
We likely need the `FILE_SET` key for `install`, which is not available for older CMake versions, so might have to split the installs up as before.
Though, I am not entirely sure how to go about installing file sets, as I've never done so before, especially with CXX modules.

CMakeLists.txt
```diff
@@ -115,7 +115,12 @@ if (VULKAN_HEADERS_ENABLE_INSTALL)
     install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/registry" DESTINATION "${CMAKE_INSTALL_DATADIR}/vulkan" USE_SOURCE_PERMISSIONS)
 
     set_target_properties(Vulkan-Headers PROPERTIES EXPORT_NAME "Headers")
-    install(TARGETS Vulkan-Headers EXPORT VulkanHeadersConfig INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+    list(APPEND INSTALL_TARGETS Vulkan-Headers)
+    if (VULKAN_HEADERS_ENABLE_MODULE)
+        set_target_properties(Vulkan-HppModule PROPERTIES EXPORT_NAME "HppModule")
+        list(APPEND INSTALL_TARGETS Vulkan-HppModule)
+    endif()
+    install(TARGETS ${INSTALL_TARGETS} EXPORT VulkanHeadersConfig INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
```
jpr42, line 123: Targets that use FILE_SET need to be installed differently.

Please look at this blog post:
https://crascit.com/2024/04/04/cxx-modules-cmake-shared-libraries/

FWIW in the future when the cmake min is higher Vulkan-Headers can be installed as a FILE_SET as well.

jpr42 (changes requested):

CMakeLists.txt
M2-TE: Thank you for linking that blog post!
I was previously missing setting a DESTINATION explicitly for the FILE_SET key.

The blog additionally mentions that CMake versions prior to 3.28.2 do not use export names for generated files.
I believe this is not necessary here, especially since we do not export multiple file sets (but could become an issue once all headers are exported as a FILE_SET as you mentioned).
Should I assert 3.28.2 for the install step? With modules enabled, the minimum is currently 3.28

jpr42 (changes requested):
vulkan.cppm is being installed twice

jpr42:
vulkan.cppm is being installed twice... It's also questionable if it should be installed in the include/vulkan directory.

```
-- Installing: /home/runner/work/Vulkan-Headers/Vulkan-Headers/build/install/include/vulkan/vulkan.cppm

...

-- Installing: /home/runner/work/Vulkan-Headers/Vulkan-Headers/build/install/lib/libVulkan-HppModule.a
-- Installing: /home/runner/work/Vulkan-Headers/Vulkan-Headers/build/install/include/include/vulkan/vulkan.cppm
-- Installing: /home/runner/work/Vulkan-Headers/Vulkan-Headers/build/install/share/cmake/VulkanHeaders/VulkanHeadersConfig.cmake
-- Installing: /home/runner/work/Vulkan-Headers/Vulkan-Headers/build/install/share/cmake/VulkanHeaders/VulkanHeadersConfig-noconfig.cmake
-- Installing: /home/runner/work/Vulkan-Headers/Vulkan-Headers/build/install/share/cmake/VulkanHeaders/./cxx-modules-VulkanHeadersConfig.cmake
-- Installing: /home/runner/work/Vulkan-Headers/Vulkan-Headers/build/install/share/cmake/VulkanHeaders/./cxx-modules-VulkanHeadersConfig-noconfig.cmake
-- Installing: /home/runner/work/Vulkan-Headers/Vulkan-Headers/build/install/share/cmake/VulkanHeaders/./target-HppModule-noconfig.cmake
```

Furthermore a libVulkan-HppModule.a is now being created as well as various *.cmake files.

What is the size of the static library? Could you size up these new artifacts on to github so we can look at them?

jpr42 (changes requested):
vulkan.cppm shouldn't be installed twice.

jpr42:
> vulkan.cppm is being installed twice

The culprit is this line of code `install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/include/vulkan" DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})`

It's installing everything even vulkan.cppm which it arguably shouldn't be... although it might break people if it gets changed now.

I'm not exactly certain with C++ modules so I can't speak with certainty.

M2-TE:
> It's installing everything even vulkan.cppm which it arguably shouldn't be... although it might break people if it gets changed now.

Yea that is what I was worried about once the normal headers are installed as file sets as well. But it is also how people get the .cppm file currently, so not putting it in the `include/vulkan/` directory would certainly break things for them.
Normally, it should be put into a `src/` folder from what I have seen, as it is basically a source file. But you could also argue that it is more akin to a header file, as we do **NOT** want to ship a compiled binary. So we basically want to ship the .cppm as a header, to be used as a source file by users.

For clarification:
> Furthermore a libVulkan-HppModule.a is now being created as well as various *.cmake files.

I cannot speak on the *.cmake files, but the behavior of creating and installing a static library is wrong.
Modules can generally be:
- Standalone files containing both declarations and implementations, akin to a header. That is what vulkan.cppm is.
- Split up into interface modules for declarations and implementation modules. This would require shipping static or shared library for the compiled implementation.

I don't know if this can be precisely controlled, as, to CMake at least, the `Vulkan-HppModule` is a library, not an `INTERFACE`. We want to ship ONLY the `Vulkan-HppModule` target, as users will want to compile these on their own. In the far future, we might additionally ship the BMI, but those are still far too unstable so not an option yet.

Out of the installed *.cmake files
```
-- Installing: /home/runner/work/Vulkan-Headers/Vulkan-Headers/build/install/share/cmake/VulkanHeaders/VulkanHeadersConfig.cmake
-- Installing: /home/runner/work/Vulkan-Headers/Vulkan-Headers/build/install/share/cmake/VulkanHeaders/VulkanHeadersConfig-noconfig.cmake
-- Installing: /home/runner/work/Vulkan-Headers/Vulkan-Headers/build/install/share/cmake/VulkanHeaders/./cxx-modules-VulkanHeadersConfig.cmake
-- Installing: /home/runner/work/Vulkan-Headers/Vulkan-Headers/build/install/share/cmake/VulkanHeaders/./cxx-modules-VulkanHeadersConfig-noconfig.cmake
-- Installing: /home/runner/work/Vulkan-Headers/Vulkan-Headers/build/install/share/cmake/VulkanHeaders/./target-HppModule-noconfig.cmake
```
we likely do not need the additional ones with the "." path specified and those probably come from the`CXX_MODULES_DIRECTORY .` key during within `install`.

The things left to do are now, as far as I can tell:
1. Remove the need to ship a binary, be it static or shared.
2. Figure out what directory the .cppm should be put into (there might be more than one in the future).
3. Assess whether the additional *.cmake files are necessary, specifically the ones with the "." path.

jpr42:
Apologies @M2-TE. Not ignoring you I just don't know enough to give a good answer. I'd recommend asking a question on the CMake discourse to clear up any confusion on what should be done here.

M2-TE:
No worries @jpr42 I meant the previous post to serve as an overview of the current state while I fight the deathly grip of some deadlines.
Let me try address some of your initial concerns again, as well as my "TODOs", after some more research:

> vulkan.cppm is being installed twice... It's also questionable if it should be installed in the include/vulkan directory.

It being installed twice is not a big issue, it will just overwrite the version that was installed initially. The odd `include/include/vulkan` path however was a mistake on my end and I've fixed the path (which was relative to the install dir), so now there will only be one .cppm.

> Furthermore a libVulkan-HppModule.a is now being created

I've opened a question on the CMake discourse as per your idea, so I will wait for feedback on that. As for it's size, its very small at around 3 KiB and is practically empty. Makes sense considering that most stuff should be inlined anyways (and the reason why I don't see the need to install it).

> as well as various *.cmake files.

These are the result of using `CXX_MODULES_DIRECTORY`, which are supposedly necessary as per cmake docs about [install](https://cmake.org/cmake/help/latest/command/install.html#export):
_Specify a subdirectory to store C++ module information for targets in the export set.
This directory will be populated with files which add the necessary target property information to the relevant targets.
Note that without this information, none of the C++ modules which are part of the targets in the export set will support being imported in consuming targets._
I did find that it install + find_package worked even without this, but I would trust the docs on this one and leave it in.

All that's left is point number 1. in my previous post about the necessity of library artifacts, for which I will simply wait for a response on the cmake discourse for now

M2-TE:
The solution seems to have been marking Vulkan-HppModule as an [OBJECT](https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries) library, therefore not shipping any build artifacts. I have not used this before, but it seems to work as far as the tests are concerned, as well as in my own local project using Vulkan-Headers.

Could you give it a quick look @jpr42 and @charles-lunarg ?

CMakeLists.txt
```diff
@@ -67,7 +67,7 @@ if (VULKAN_HEADERS_ENABLE_MODULE)
         endif()
         
         # set up Vulkan-HppModule
-        add_library(Vulkan-HppModule)
+        add_library(Vulkan-HppModule OBJECT)
```
jpr42, line 70: Could this be made into a pure interface library like Vulkan-Headers?

M2-TE: Sadly no, targets with C++20-module sources cannot be made interface:
> [cmake] target_sources File set TYPE "CXX_MODULES" may not have "INTERFACE" visibility

It's the conundrum of being able to treat .cppm as both a header and a source file

jpr42 (approved):

charles-lunarg (approved):

PR #540, opened 2025-04-13
ignore Jetbrains config directories

brccabral:
Basic Jetbrains (CLion) setup creates .idea/, cmake-build-debug/, cmake-build-release/ directories.


charles-lunarg (approved):
Works for me - the gitignore is for developer convenience, no reason to be selective about the IDE's that are added to it.

PR #555, opened 2025-07-17
Remove VVL workaround

artem-lunarg:
Fixes validation of host stage in Vulkan-ValidationLayers.

The initial intention of the workaround was to fix a crash with None value, but it also introduced side effect when HOST_STAGE is considered to be part of TRANSFER_QUEUE, which led to incorrect validation of HOST_STAGE.

The None value (no queue) in VVL scripts appears as the result that HOST_STAGE is not assigned to any queue. VVL can handle this with explicit check without using special values.


oddhack:
You'll need to retarget this to Vulkan-Docs (or preferably, internal github). This repo is only a publishing mechanism.

PR #557, opened 2025-07-31
CMake: Don't install C++ module headers if module build is disabled

sarnex:
Currently we are installing `.cppm` files even if `VULKAN_HEADERS_ENABLE_MODULE` is off.

Right now in Gentoo we are just manually deleting the files, would be nice to have a clean solution.

Let me know if there's a way to prevent the copy paste here.


charles-lunarg:
Agreed. If you want module support it must be opt-in, install should respect that.

sarnex:
Thanks Charles!

PR #558, opened 2025-08-01
Update for Vulkan-Docs 1.4.324

oddhack:


oddhack:
@asuessenbach could you take a look at the CI failure in https://github.com/KhronosGroup/Vulkan-Headers/pull/558 ? It appears something is broken in the updated Vulkan-Hpp headers that we'll need help with.

asuessenbach:
How would the regeneration of the Vulkan-Hpp files be triggered?
I think, I fixed this issue, but don't get updated Vulkan-Hpp files in here.

oddhack:
> How would the regeneration of the Vulkan-Hpp files be triggered? I think, I fixed this issue, but don't get updated Vulkan-Hpp files in here.

I regenerated them from Vulkan-Docs. Everything looks OK now so I will proceed with updating Vulkan-Headers - thanks!

PR #567, opened 2025-09-25
CI: Fixing workflow for C++20 modules

### LINKED ISSUE #566: v1.4.328 update failing in cmake-windows-modules CI job
It is passing all other CI stages but failing in e.g. https://github.com/KhronosGroup/Vulkan-Headers/actions/runs/18021294685/job/51279254944?pr=565 . I don't know what this indicates - there are so many different compilers and language versions involved in CI. Perhaps Vulkan-Hpp should have a comparable test matrix in its CI if it does not already, so problems at that end can be turned up more rapidly.

@charles-lunarg should I go ahead and merge this so the new extension is in Vulkan-Headers, even if something about CI or vulkan.hpp is broken with respect to this specific compiler path?

I can't diagnose this very far, but re-ran the previous v1.4.327 cmake-windows-modules job and it still passed, so conceivably something in Vulkan-Hpp is reacting to XML changes in .328 to generate some construct illegal on this compiler path? Also there's a warning at the bottom of the job about github moving from Window Server 2022 -> 2025 if that might have any relevance.

M2-TE:
Resolves #566.

As described in the issue above, CI toolchains are not yet set up to handle `import std`, which gets used automatically, when the compiler is (theoretically) capable of it. A flag `VULKAN_HEADERS_DISABLE_MODULE_STD` was added here to toggle `VULKAN_HPP_NO_STD_MODULE` to disable this behavior.

`VULKAN_HEADERS_ENABLE_MODULE_STD` can likely be phased out, but we are still actively looking into how to best handle the whole import std situation, so that should wait for now.

(sorry for the spotty commits, that's the midnight tax right there)


CMakeLists.txt
```diff
@@ -43,6 +43,7 @@ option(VULKAN_HEADERS_ENABLE_TESTS "Test Vulkan-Headers" ${PROJECT_IS_TOP_LEVEL}
 option(VULKAN_HEADERS_ENABLE_INSTALL "Install Vulkan-Headers" ${PROJECT_IS_TOP_LEVEL})
 option(VULKAN_HEADERS_ENABLE_MODULE "Enables building of the Vulkan C++20 module; requires minimum CMake version 3.28" OFF)
 option(VULKAN_HEADERS_ENABLE_MODULE_STD "Enables building of the Vulkan C++20 module with import std; requires minimum CMake version 3.30" OFF)
+option(VULKAN_HEADERS_DISABLE_MODULE_STD "Disable automatic use of import std; for the Vulkan C++20 module" OFF)
```
charles-lunarg, line 46: This makes it rather confusing for users. Two options, one to enable it, one to disable it? 

Is there a reason why `VULKAN_HPP_NO_STD_MODULE` be set to the inverse of `VULKAN_HEADERS_ENABLE_MODULE_STD`?

charles-lunarg: Alternatively, just replace `VULKAN_HEADERS_ENABLE_MODULE_STD` with `VULKAN_HEADERS_DISABLE_MODULE_STD` as the former is unique to this repo. If using module_std is going to be the default going forward, it makes sense to change the option to an 'opt-out' rather than a 'opt-in' that it is currently.

M2-TE: A better naming for the enabling flag would perhaps be `VULKAN_HEADERS_FORCE_MODULE_STD`?

charles-lunarg: Perhaps but renaming options is generally more messy than helpful. The current names are good, and yes I agree the ENABLE option will go away as the issues are ironed out.

charles-lunarg:
The module code is off by default. Does that suffice to prevent breaking downstream users?

sharadhr:
I have moved discussion to #568.

PR #611, opened 2026-06-26
Change to dual license on static Vulkan headers

### LINKED ISSUE #605: License inconsistency in headers
Some headers list only list Apache-2.0, while some list Apache-2.0 OR MIT:
https://github.com/KhronosGroup/Vulkan-Headers/blob/0e9de566b7d4051c5cc1b762e242c46565956bdf/include/vulkan/vulkan.h#L7
`** SPDX-License-Identifier: Apache-2.0`

but vulkan.hpp says both:
https://github.com/KhronosGroup/Vulkan-Headers/blob/0e9de566b7d4051c5cc1b762e242c46565956bdf/include/vulkan/vulkan.hpp#L3
`// SPDX-License-Identifier: Apache-2.0 OR MIT`

This is causing some discussion in GPLv2 projects, where MIT is compatible, but Apache-2.0 is not. I was wondering if we could have some clarification whether certain files are only intended for Apache-2.0, and not for MIT, and if not, could the headers be updated with `OR MIT`.

Thank you!

oddhack:
Based on feedback here and internal Vulkan issue 4883, changing the license on the static headers vk_icd.h and vk_layer.h to dual Apache-2.0 OR MIT, matching the generated headers.

Closes #605

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,12 +1,13 @@
-Copyright 2015-2023 The Khronos Group Inc.
+Copyright 2015-2026 The Khronos Group Inc.
 
 Files in this repository fall under one of these licenses:
 
 - `Apache-2.0`
 - `MIT`
 
-Note: With the exception of `parse_dependency.py` the files using `MIT` license
-also fall under `Apache-2.0`. Example:
+Note: With the exception of `parse_dependency.py`, which is based on an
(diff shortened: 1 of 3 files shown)


charles-lunarg (approved):

oddhack:
Waiting for Ralph to sign off in the internal issue as the WG chair before merging.

PR #613, opened 2026-07-07
Update CMake min to 3.28

jpr42:
Ubuntu 24.04 LTS (noble) packages cmake 3.28.3, only the 2 latest LTS releases are supported,
requiring 3.28 keeps compatibility with all currently supported Ubuntu LTS versions.

Utilize new block() functionality instead of function

Utilize VERIFY_INTERFACE_HEADER_SETS to ensure header files can compile correctly on their own. (Test code only).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,16 +5,17 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 # ~~~
-cmake_minimum_required(VERSION 3.22.1)
+cmake_minimum_required(VERSION 3.28.3)
 
 include(CMakeDependentOption)
-
 
 # NOTE: Parsing the version like this is suboptimal but neccessary due to our release process:
 # https://github.com/KhronosGroup/Vulkan-Headers/pull/346
 #
-# As shown a more robust approach would be just to add basic test code to check the project version.
-function(vlk_get_header_version)
+# A better approach would be just to add basic test code to check the project version.
+# IE set the version in the CMakeLists.txt and then check that the version is correct in a test.
+# Avoids file parsing on every single configuration and is more robust to changes in the header file.
+block(PROPAGATE VK_VERSION_STRING)
     set(vulkan_core_header_file "${CMAKE_CURRENT_SOURCE_DIR}/include/vulkan/vulkan_core.h")
     if (NOT EXISTS ${vulkan_core_header_file})
         message(FATAL_ERROR "Couldn't find vulkan_core.h!")
@@ -35,9 +36,8 @@ function(vlk_get_header_version)
         message(FATAL_ERROR "Couldn't get the patch version")
     endif()
 
-    set(VK_VERSION_STRING "${MAJOR_VERSION}.${MINOR_VERSION}.${PATCH_VERSION}" PARENT_SCOPE)
-endfunction()
-vlk_get_header_version()
+    set(VK_VERSION_STRING "${MAJOR_VERSION}.${MINOR_VERSION}.${PATCH_VERSION}")
+endblock()
 
 project(VULKAN_HEADERS LANGUAGES C CXX VERSION ${VK_VERSION_STRING})
 
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -34,3 +34,70 @@ add_test(NAME integration.find_package
 
 # Installing comes before testing
 set_tests_properties(integration.find_package PROPERTIES DEPENDS integration.install)
+
+# Verify that every public header compiles standalone (i.e. that each one includes
+# everything it needs and doesn't just rely on being pulled in after some other header).
+#
+# The header list is globbed so newly added headers are picked up automatically. Headers
+# that are known not to be self-contained are excluded from compilation individually via
+# SKIP_LINTING below, rather than omitted from the FILE_SET, so they still show up here
+# and don't silently fall out of the verification as the headers evolve.
+file(GLOB_RECURSE vk_headers_verify_headers CONFIGURE_DEPENDS LIST_DIRECTORIES false
+    "${VULKAN_HEADERS_SOURCE_DIR}/include/vulkan/*.h"
+    "${VULKAN_HEADERS_SOURCE_DIR}/include/vulkan/*.hpp"
+    "${VULKAN_HEADERS_SOURCE_DIR}/include/vk_video/*.h"
+)
+list(SORT vk_headers_verify_headers)
+
+add_library(vk_headers_verify INTERFACE)
+set_target_properties(vk_headers_verify PROPERTIES LINKER_LANGUAGE C)
+target_sources(vk_headers_verify INTERFACE
+    FILE_SET HEADERS
+    BASE_DIRS "${VULKAN_HEADERS_SOURCE_DIR}/include"
+    FILES ${vk_headers_verify_headers}
+)
+
+# WSI headers (vulkan_win32.h, vulkan_xcb.h, vulkan_wayland.h, ...) are deliberately
+# skipped: vulkan.h itself includes their platform prerequisites (e.g. <windows.h>) before
+# including them, so they're sub-headers of vulkan.h, not standalone headers.
+set(vk_headers_verify_platform_headers
+    vulkan_android.h
+    vulkan_directfb.h
+    vulkan_fuchsia.h
+    vulkan_ggp.h
+    vulkan_ios.h
+    vulkan_macos.h
+    vulkan_metal.h
+    vulkan_ohos.h
+    vulkan_screen.h
+    vulkan_ubm.h
+    vulkan_vi.h
+    vulkan_wayland.h
+    vulkan_win32.h
+    vulkan_xcb.h
+    vulkan_xlib.h
+    vulkan_xlib_xrandr.h
+)
+
+# These files cannot be included on their own.
+set(vk_headers_verify_detail_headers
+    vulkan_beta.h
+    vulkan_enums.hpp
+    vulkan_funcs.hpp
+    vulkan_handles.hpp
+    vulkan_structs.hpp
+    vulkan_to_string.hpp
+)
+
+foreach(header IN LISTS vk_headers_verify_platform_headers vk_headers_verify_detail_headers)
+    set_source_files_properties("${VULKAN_HEADERS_SOURCE_DIR}/include/vulkan/${header}" PROPERTIES SKIP_LINTING ON)
+endforeach()
+
+set_target_properties(vk_headers_verify PROPERTIES VERIFY_INTERFACE_HEADER_SETS ON)
+
+# CMake always sets EXCLUDE_FROM_ALL on the generated all_verify_interface_header_sets
+# target, so a plain build never compiles it; this explicit --target build is the only
+# way to run the verification.
+add_test(NAME headers.verify_standalone
+    COMMAND ${CMAKE_COMMAND} --build ${VULKAN_HEADERS_BINARY_DIR} --target all_verify_interface_header_sets --config $<CONFIG>
+)
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -38,23 +38,31 @@ set_tests_properties(integration.find_package PROPERTIES DEPENDS integration.ins
 # Verify that every public header compiles standalone (i.e. that each one includes
 # everything it needs and doesn't just rely on being pulled in after some other header).
 #
-# The header list is globbed so newly added headers are picked up automatically. Headers
-# that are known not to be self-contained are excluded from compilation individually via
-# SKIP_LINTING below, rather than omitted from the FILE_SET, so they still show up here
-# and don't silently fall out of the verification as the headers evolve.
-file(GLOB_RECURSE vk_headers_verify_headers CONFIGURE_DEPENDS LIST_DIRECTORIES false
-    "${VULKAN_HEADERS_SOURCE_DIR}/include/vulkan/*.h"
-    "${VULKAN_HEADERS_SOURCE_DIR}/include/vulkan/*.hpp"
-    "${VULKAN_HEADERS_SOURCE_DIR}/include/vk_video/*.h"
+# The header list is globbed so newly added headers are picked up automatically, in any
+# subdirectory of include/. Headers that are known not to be self-contained are excluded
+# from compilation individually via SKIP_LINTING below, rather than omitted from the
+# FILE_SET, so they still show up here and don't silently fall out of the verification as
+# the headers evolve.
+file(GLOB_RECURSE vk_headers_verify_c_headers CONFIGURE_DEPENDS LIST_DIRECTORIES false
+    "${VULKAN_HEADERS_SOURCE_DIR}/include/*.h"
 )
-list(SORT vk_headers_verify_headers)
+file(GLOB_RECURSE vk_headers_verify_cxx_headers CONFIGURE_DEPENDS LIST_DIRECTORIES false
+    "${VULKAN_HEADERS_SOURCE_DIR}/include/*.hpp"
+)
+list(SORT vk_headers_verify_c_headers)
+list(SORT vk_headers_verify_cxx_headers)
+
+# Set LANGUAGE explicitly per header instead of leaving it for CMake to deduce: .h is an
+# extension both C and CXX claim, so an undeduced .h header could silently end up compiled
+# as C++ instead of being checked against a C compiler.
+set_source_files_properties(${vk_headers_verify_c_headers} PROPERTIES LANGUAGE C)
+set_source_files_properties(${vk_headers_verify_cxx_headers} PROPERTIES LANGUAGE CXX)
 
 add_library(vk_headers_verify INTERFACE)
-set_target_properties(vk_headers_verify PROPERTIES LINKER_LANGUAGE C)
 target_sources(vk_headers_verify INTERFACE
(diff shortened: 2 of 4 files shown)


tests/CMakeLists.txt
```diff
@@ -34,3 +34,70 @@ add_test(NAME integration.find_package
...
+
+# These files cannot be included on their own.
+set(vk_headers_verify_detail_headers
+    vulkan_beta.h
+    vulkan_enums.hpp
+    vulkan_funcs.hpp
+    vulkan_handles.hpp
+    vulkan_structs.hpp
+    vulkan_to_string.hpp
+)
```
jpr42: @charles-lunarg should these files be standalone compile-able?

M2-TE: Speaking only for the `*.hpp` files, they aren't designed to be included on their own. Though, they get pulled in automatically by including `vulkan.hpp`, so there should not be any need to test them individually.

A list of what other vulkan-related headers `vulkan.hpp` includes by default:
```cpp
#include <vulkan/vulkan.h>
#include <vulkan/vulkan_hpp_macros.hpp>
#include <vulkan/vulkan_enums.hpp>
#include <vulkan/vulkan_to_string.hpp>
#include <vulkan/vulkan_handles.hpp>
#include <vulkan/vulkan_structs.hpp>
#include <vulkan/vulkan_funcs.hpp>
```

jpr42: Thanks for the feedback. I'll update the new test accordingly.

tests/CMakeLists.txt
```diff
@@ -34,3 +34,80 @@ add_test(NAME integration.find_package
...
+# and don't silently fall out of the verification as the headers evolve.
+file(GLOB_RECURSE vk_headers_verify_headers CONFIGURE_DEPENDS LIST_DIRECTORIES false
+    "${VULKAN_HEADERS_SOURCE_DIR}/include/vulkan/*.h"
+    "${VULKAN_HEADERS_SOURCE_DIR}/include/vulkan/*.hpp"
+    "${VULKAN_HEADERS_SOURCE_DIR}/include/vk_video/*.h"
+)
+list(SORT vk_headers_verify_headers)
+
+add_library(vk_headers_verify INTERFACE)
+set_target_properties(vk_headers_verify PROPERTIES LINKER_LANGUAGE C)
```
M2-TE: It does not seem to pose a problem, but I wanted to draw attention to the fact that you assign the C [LINKER_LANGUAGE](https://cmake.org/cmake/help/latest/prop_tgt/LINKER_LANGUAGE.html), even though there are some cpp headers sprinkled in here.

For the  linker, the only difference would be the mangled names in cpp. Does this also control which compiler is invoked for each header verification? Or does CMake automatically use the `CMAKE_CXX_COMPILER` for `.hpp` files? As per [VERIFY_INTERFACE_HEADER_SETS](https://cmake.org/cmake/help/latest/prop_tgt/VERIFY_INTERFACE_HEADER_SETS.html#prop_tgt:VERIFY_INTERFACE_HEADER_SETS):

> If the header's [LANGUAGE](https://cmake.org/cmake/help/latest/prop_sf/LANGUAGE.html#prop_sf:LANGUAGE) property is set, the value of that property is used to determine the language with which to compile the header file. The supported languages are C, CXX, OBJC, and OBJCXX. If the language is not one of those four, the header file is not compiled.

When [LANGUAGE](https://cmake.org/cmake/help/latest/prop_sf/LANGUAGE.html#prop_sf:LANGUAGE) is not set for a file, it will automatically deduce the compiler, _but_ these are headers and not source files. I am not certain that e.g. `.hpp` will have a compiler assigned to it; but then again, the test seems to run just fine and only exists to check for included header completeness.

Perhaps too paranoid, but an option could be two separate `GLOB_RECURSE` functions, one for `.h` and another for `.hpp`, for each of which a specific `LANGUAGE` can be set to the resulting file lists.

tests/CMakeLists.txt
```diff
@@ -34,3 +34,80 @@ add_test(NAME integration.find_package
...
+# everything it needs and doesn't just rely on being pulled in after some other header).
+#
+# The header list is globbed so newly added headers are picked up automatically. Headers
+# that are known not to be self-contained are excluded from compilation individually via
+# SKIP_LINTING below, rather than omitted from the FILE_SET, so they still show up here
+# and don't silently fall out of the verification as the headers evolve.
+file(GLOB_RECURSE vk_headers_verify_headers CONFIGURE_DEPENDS LIST_DIRECTORIES false
+    "${VULKAN_HEADERS_SOURCE_DIR}/include/vulkan/*.h"
+    "${VULKAN_HEADERS_SOURCE_DIR}/include/vulkan/*.hpp"
+    "${VULKAN_HEADERS_SOURCE_DIR}/include/vk_video/*.h"
```
M2-TE:
```suggestion
    "${VULKAN_HEADERS_SOURCE_DIR}/include/*.h"
    "${VULKAN_HEADERS_SOURCE_DIR}/include/*.hpp"
```

For robustness in case another folder gets added in the future, these can be simplified to look for _all_ headers in `include/*`. It already uses the recursive glob anyways.

M2-TE (commented):
Should already be good in its current state.
Still, I've added two minor comments below.

Another thing: The windows clang runners only seem to use the ~latest~ default cmake, not `[ '3.22.1', 'latest']` like the others. Is this intentional?

tests/CMakeLists.txt
jpr42: Fixed.

jpr42:
> Another thing: The windows clang runners only seem to use the ~latest~ default cmake, not `[ '3.22.1', 'latest']` like the others. Is this intentional?

Not sure. I fixed it regardless.

M2-TE:
Looks good. Got nothing else to add!

charles-lunarg (approved):
Wow it felt like CMake 3.28 released last week... it came out at the tail end of 2023. 

Looking through various Vulkan repo's, the min cmake version is 3.22.1, which stems from an update to spirv-tools that occured in Jan 2025. 

I am going to approve this PR because it is in a good state, and moves the baseline forward, but I am holding off merging until I can get a confirmation that it is too soon to update. Merging this will likely require dependent repo's to update their CMake version, so in a sense this triggers a small cascade of subsequent PR's in about a dozen repo's.

jpr42:
@spencer-lunarg can this be merged in now?

Last I heard about this from @charles-lunarg was that LunarG hadn't updated all the CI machines with CMake 3.28+ yet.

spencer-lunarg (approved):

CMakeLists.txt
```diff
@@ -5,16 +5,17 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 # ~~~
-cmake_minimum_required(VERSION 3.22.1)
+cmake_minimum_required(VERSION 3.28.3)
```
spencer-lunarg, line 8: actually @jpr42 is there a "need" to update this to 3.28, like are the various new additions below making use of it or this purely to align things to Ubuntu 24.04?

jpr42: I did add new functionality that relies on a new version of CMake. I also have plans to update the loader/vvl with various CMake updates as well

spencer-lunarg: so I am going to bring this up in the Working Group next week (to make sure everyone knows this is coming)... to prepare for someone to go "why risk breaking things, what does this buy us to update" you have a response ready for me

jpr42: We did this the last time a new Ubuntu LTS came out. There is already precedent.

tests/CMakeLists.txt
```diff
@@ -34,3 +34,88 @@ add_test(NAME integration.find_package
...
+target_sources(vk_headers_verify INTERFACE
+    FILE_SET HEADERS
+    BASE_DIRS "${VULKAN_HEADERS_SOURCE_DIR}/include"
+    FILES ${vk_headers_verify_c_headers} ${vk_headers_verify_cxx_headers}
+)
+
+# WSI headers (vulkan_win32.h, vulkan_xcb.h, vulkan_wayland.h, ...) are deliberately
+# skipped: vulkan.h itself includes their platform prerequisites (e.g. <windows.h>) before
+# including them, so they're sub-headers of vulkan.h, not standalone headers.
+set(vk_headers_verify_platform_headers
```
oddhack, line 71: I wonder if this information could be extracted out of vulkan.h? Otherwise when a new platform is added, someone will probably need to know to make the update here as well.

oddhack:
@spencer-lunarg had asked to discuss this on the maintenance call next week because I would be there, but I don't have much of an opinion on this - I just update the headers when a spec update is done, while all the CI infrastructure here is provided by LunarG AFAIK. I did make one minor comment about the explicit header list added to a cmake file but from my perspective this PR is fine.

spencer-lunarg:
Update @jpr42 - So the working group was not excited to raise the CMake version purely for the sake of it

> There is already precedent

People also didn't agree to this and felt that CMake should be the lowest possible version for the headers as many tools grab it with older toolchains

Hans K pointed out that updating this would add churn to some internal Steam tools and without a solid reason for the churn is not incline to take this PR

> only the 2 latest LTS releases are supported,

This is a LunarG claim, not everyone using the headers claim this

M2-TE:
Could we not just use a version range with [cmake_minimum_required](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html)?
It keeps compatibility with old versions, but allows using features from newer ones.

e.g.:

```cmake
cmake_minimum_required(VERSION 3.22.1...3.28.3)
```

That likely would not require any changes anywhere outside this repo and allows seamlessly updating to and/or testing newer cmake features. Just need a version check within cmake code for code sections like the things in this pr.

jpr42:
Totally fair, and thanks for taking this to the WG — happy to leave this PR closed, not trying to push it through as-is.

A couple of things worth leaving on the table for next time this comes up, with sources rather than just my opinion.

This isn't unprecedented within Khronos itself. SPIRV-Tools has bumped its own CMake minimum twice — 2.18.2→3.17.2 ([issue #5040](https://github.com/KhronosGroup/SPIRV-Tools/issues/5040), 2022) and 3.17.2→3.22.1 ([PR #5925](https://github.com/KhronosGroup/SPIRV-Tools/pull/5925), 2024). The rationale for the first bump is worth quoting directly: "I'd like to update minimum cmake to 3.17.2, the max among our tested configurations" — i.e. set the floor to what CI already exercises, not to whatever's oldest and untested. Worth noting where that's left things: Vulkan-Loader, Vulkan-ValidationLayers, Vulkan-Utility-Libraries, glslang, Vulkan-Tools, Vulkan-Profiles, Vulkan-ExtensionLayer, Vulkan-Hpp, and SPIRV-Tools have all already converged on CMake 3.22/3.22.1. This kind of bump isn't unusual for this ecosystem, it's already happened repeatedly, on roughly this cadence.

Worth putting next to each other: VVL and Vulkan-Utility-Libraries already require C++17 today, and C++20 is reportedly next — both real toolchain-currency asks. C++17 alone already rules out compilers that aren't that old — full support landed in GCC 8, Clang 5, and VS2017 update 15.8. A CMake minimum bump is the same category of ask, just aimed at a different part of the toolchain. LLVM's own history is a useful data point here: they don't treat any one piece of the toolchain as permanently frozen while letting the others move. They've revisited their minimum Python version on basically the same cadence as CMake — an EOL-driven bump in 2023, another in 2025, and a follow-up proposal to put Python bumps on a formal annual schedule specifically so it stops being re-litigated from scratch every time. In fairness, one commenter on that thread pushed back that compiler-vs-Python isn't a perfect comparison — compilers are more central and more heterogeneous — which is fair. But the broader pattern still holds: none of compiler, Python, or CMake gets a permanent "never touch this one" exemption over there; each gets revisited on its own merits as needs and support windows change.

C++20 modules specifically need a fairly new CMake, and it's still catching up release-to-release. Named modules only left experimental status in CMake 3.28 (Kitware: "import CMake; the Experiment is Over!"), and real bug fixes kept landing through 3.28.2 and beyond. import std; for C++23 didn't arrive until 3.30 (funded by Bloomberg) and is still gated experimental there (Kitware: "import std in CMake 3.30"). Even past 3.28, shared-library module support has real generator-specific gaps depending on which release you're on (Crascit: C++20 Modules, CMake, and Shared Libraries). If modules are part of where these repos are headed, "keep CMake as old as possible" and "adopt modules" pull in opposite directions — they're not independent choices.

This same conversation has played out repeatedly outside Khronos too, and resolves the same way each time. LLVM has had this debate in public more than once: 2020, 2022 (→ became official at 3.20.0), and twice already this year (RFC for 3.28.3/3.30, follow-up RFC for 3.31). Boost had a version of this same conversation too — a [2021 mailing-list thread](https://groups.google.com/g/boost-developers-archive/c/oQyTNm87VLg) weighed holding the floor at 3.8/3.9 for RHEL7-era compatibility against moving up for better Python/install support, and opinion leaned toward the newer end (3.13–3.15) rather than staying frozen at the RHEL7 floor. Worth being honest that not every stated minimum across the ecosystem is a carefully-chosen line — a lot are just copy-pasted and never rechecked (relevant Lobsters discussion) — but that cuts in favor of periodically re-verifying the floor against something concrete (tested CI configs, needed features), same as SPIRV-Tools did, rather than leaving it untouched by default.

One forward-looking reason not to let the floor freeze at 3.22: Kitware's Common Package Specification (CPS) — a vendor-neutral package-description format meant to let CMake and other build/dependency tools exchange package info without hand-rolled Find*.cmake modules — went from experimental export in CMake 3.31, to experimental import in 4.0, to fully stable in 4.3 (Kitware: "Common Package Specification is Out the Gate", background). If Vulkan's CMake packages ever want in on that, that's another force pulling the realistic floor forward.

Not trying to relitigate this specific PR — just wanted the primary sources on record for whenever this comes up again.

My main goal with all of this is to make Vulkan software robust, modern, and easy to contribute to.

> Could we not just use a version range with [cmake_minimum_required](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html)?

Thanks — that's a good instinct, and worth clarifying since cmake_minimum_required(VERSION <min>...<max>) is genuinely confusing.

That range syntax only controls policy behavior, not command/feature availability. Per the [CMake docs](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html): it sets policies up through <max> to their NEW behavior if the CMake actually running the script is that new, but it doesn't backport anything to older CMake. <min> is still the real floor for what has to be installed to build at all.

Concretely, for this PR's diff: it uses block()/endblock() (added in CMake 3.25) and FILE_SET HEADERS / VERIFY_INTERFACE_HEADER_SETS (added in 3.23/3.24) unconditionally in tests/CMakeLists.txt. Those are commands that don't exist at all in 3.22.1 — not a policy difference, an actual missing command — so cmake_minimum_required(VERSION 3.22.1...3.28.3) wouldn't help here. Anyone on 3.22.1 would still hit "unknown command" the moment CMake parses those lines, regardless of the range.

To get what you're describing (old floor, new features opt-in), you'd need explicit version guards around each new-command call site — if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.25) ... block() ... endif() and similar for the FILE_SET usage — rather than relying on the range syntax to do it implicitly. That's a real option, but it comes with its own cost: every version-gated code path is something the project has to maintain and test in both branches indefinitely, which is part of what motivated bumping the floor outright instead.

Good thing to have on record either way — this distinction between "policy compatibility range" and "minimum version needed to run the commands you actually use" trips people up a lot, and it'll matter again whenever this comes up for Vulkan-Headers or its sibling repos.

dneto0:
I appreciate you writing down detailed notes.

I'm responsible for some of the past decisions and actions, e.g. updating `cmake_minimum_required` on SPIRV-Tools and some other repositories.

Very briefly, about setting `cmake_minimum_required`:

- I want to support many downstream developers, with reasonable maintenance costs, and high stability and security.
     - So I look at what is available in LTS distributions.
- I updated SPIRV-Tools cmake_minimum_required to 3.22.1 around Christmas 2024.   I chose 3.22.1 because it aligned with Ubuntu 22.04 LTS.
- Broadly, I'm also guided by Google's recommended minimums for open source projects. See  https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md  
    -  That currently recommends a minimum of `>= 3.22`, with the rational being "We support the oldest version of CMake that ships with one of the supported distros."
    -  They also say they will change the recommended minimum in spring 2027 when Ubuntu 22.04 LTS goes end-of-life.

Pulling all that together, I agree with the decision not to push the `cmake_minimum_required` forward at this time.  It also sets an expectation for when to do it next.


Separately, but not quite orthogonally, what tools to test with:
- My main concern is to ensure repos _test_ with the same versions of tools.  This simplifies maintenance and upgrades.
- Concretely, I maintain the Kokoro-based CI bots on 7 repos/projects in this ecosystem. 
- I try to keep tool versions in sync across those 7 repos.   Currently they use CMake 3.31.2 which was the latest stable release when I did the update.
- I do rounds of updates every once in a while, for each of about 12 different tools across repos and operating systems.

This is relevant because:
- Upgrades are costly: it takes time to do them, and they introduce risk.
- As I update CMake, older configurations are untested and may bit-rot.  The bigger the gap between the `cmake_minimum_required` and the actually tested configuration, the bigger the risk.
- Also, CMake deprecates and deletes support for features over time.  That means we can't stop the clock forever.

So we have to remain thoughtful on how we set minimums, with various stakeholders and costs taken into account.

